PR #18, opened 2026-06-26
Fix DuckDB connection failure with coexisting connections

skaltman:
From Claude (have tested, but not verified in real app). Putting this here in case you have a chance to take a look, otherwise I will test it out later. 

## Problem

`duckdb_connect()` passed `home_directory` in the DuckDB `config`. But `home_directory` is a process-global, set-at-startup option, so re-setting it on a second connection in the same process fails with:

> Could not set option "home_directory" as a global option

This happens whenever two DuckDB connections coexist — e.g. the connection `query_measure_output()` opens for SQL post-processing while a data-frame `data_source()` still holds one. The result is that `call_measure(..., sql = ...)` can fail in a long-lived Shiny/Connect app. (On duckdb 1.3.2 the config-with-`home_directory` connect fails on the *first* connection too.)

## Fix

- Keep `extension_directory` in `config` — it must be set at startup, before the extension subsystem initializes against the read-only package library on Connect.
- Set `home_directory` with a session-scoped `SET home_directory = ...` after connecting, which avoids the global-option conflict and is safe across coexisting connections.

image

### DIFF
--- a/R/data-source.R
+++ b/R/data-source.R
@@ -227,15 +227,26 @@ SET lock_configuration = true;
 
 # DuckDB defaults its extension and home directories to the package library,
 # which is read-only in deployed environments (e.g. Posit Connect). Point them
-# at the session temp directory. This must happen at connection time: any query
-# (even `SET extension_directory`) first initializes the extension subsystem
-# against the default directory, which fails when that directory is read-only.
+# at the session temp directory.
+#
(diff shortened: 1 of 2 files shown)


simonpcouch (approved):

skaltman:
Tested this in tiles-agent and it appears to be working.

PR #20, opened 2026-07-06
Fix scope trap, support multiple data sources

### LINKED ISSUE #19: The "scope trap" is unfortunate
We don't need to address this right now, but if commons' trajectory heads towards public release, I'd like to carefully weigh the different solutions to this problem.

simonpcouch:
Closes #19.

`commons(data_source)` is now `commons(data_sources)` and accepts a named list of `data_source()`s. If there are many, their name prefixes the selected tables as a header in the system prompt. For `run_sql` and `describe_table`, the model supplies the name of the connection from an enum (measures handle this automatically).

Non-`data_source()` data source should use a normal argument default in the relevant measure's signature.


simonpcouch:
> Other non-data_source() arguments are automatically provided to measure functions by name.

These should just be normal argument defaults (recommended to be a function) since `run_sql` / `describe_table` doesn't need them.

PR #22, opened 2026-07-08
Support data dictionaries

### LINKED ISSUE #21: Entry point for table- and column-level context
commons currently doesn't have a natural, first-class way of attaching context about specific data sources or data source columns to the system.

Several platforms have definitions of data dictionaries or portions of the semantic layer that reasonably approximate a data dictionary. There's also the OSS spec data-dict.yaml: https://github.com/tidyverse/data-dict

A few thoughts on integrating data dictionary-style context sources into commons...

* When should the agent "see" the data dictionary information? We want something that 1) results only in the agent seeing the necessary information (i.e. the data dictionary elements for the queried table and column) and not more (token-efficiency), 2) the agent being very likely to see the information it ought to, 3) does not increase time-to-answer if it doesn't have to. Should this be a new tool? Or injected as part of run_sql results that touch a given column / table? I don't know that pushing it into (only) `context_layer()` is what we want.
* How do data dictionaries get ingested? How different are the various things that are similarly shaped to data-dict.yaml and how machine-reable are they? data-dict.yaml is probably the framework that we will push users to use by default if they do not have such a context source already, but I would anticipate many do already.

simonpcouch:
Closes #21. `data_source()` gains a `dictionary` argument that, for now, is a path to a data dictionary. From the docs:

> Its content reaches the agent three ways:
> 
> * The dataset-level `description` and `details`, along with the glossary, are included in the system prompt. These fields are the place for rules that span tables and for guidance on which tables answer which kinds of questions.
> * The first time a conversation touches a table---via the `describe_table` tool or a SQL query---the table's full dictionary entry rides along with the tool result: its prose, documented columns, relationships, and definitions of glossary terms it references. `describe_table` merges documented columns with the table's live schema.
> * When the agent also has a `context_layer()`, the dictionary's prose is indexed for the `search_context` tool.


simonpcouch:
Merging for now but very much open to reworking after we get a sense for what tiles-agent might actually need.

PR #24, opened 2026-07-10
 Control where trajectories are logged and who can see them

skaltman:
I know the pins are probably an intermediate solution, but there wasn't a good way to share with them with others. I wanted to share all trajectories from the tiles agent with you and Kelsey. 

The `log` argument of `commons()` now accepts a `log_*()` function in
addition to the existing `TRUE`/`FALSE`/string shorthands:

- `log_local(path)`: write trajectory files to a local directory.
- `log_pins(share_with = NULL)`: write private Connect pins, and optionally
  grant named Connect users viewer access to each conversation's pin.

```r
# Log to private Connect pins, shared with two colleagues
commons(..., log = log_pins(share_with = c("simonpcouch@posit.co")))
```
`TRUE` still auto-selects and `FALSE` still disables logging.


simonpcouch (approved):
Nice, let's do it!

Heads up that you're right that this was intended as an intermediate solution and I'll be meeting with Aaron later today to talk through the "real" solution. Will be sure to loop you in when this changes.

PR #27, opened 2026-07-10
build the context store lazily and prewarm it from the Shiny module

simonpcouch:
cc @skaltman—this schedules the building of the context layer for the moment the UI appears, so the page will load much more quickly but the server will be busy for the same amount of time that it previously took the page to load. This should be an issue unless users send a message relatively quickly on app load.

This will deserve a more thorough treatment later.


skaltman:
It looks like this helps speed things up, so I'm going to merge!

PR #31, opened 2026-07-20
rewrite trajectory logging on OpenTelemetry

simonpcouch:
> Replaces the pins-based trajectory machinery with Connect's native OTel observability: ellmer's GenAI spans carry the message history, commons stamps each turn with a wrapper span carrying gen_ai.conversation.id, and read_trajectories() rebuilds ellmer turns from OTLP spans fetched from Connect's traces API or read from local NDJSON files. log = TRUE is the whole opt-in, configuring a local file exporter automatically when no exporter is set up; share_with now grants collaborator access to the deployed content, since reading traces requires editor-level permission.


R/tracing.R
```diff
@@ -0,0 +1,228 @@
...
+# function. If ellmer's internals change, capture silently stays off for the
+# session; setting the env vars before R starts (as `warn_tracing_disabled()`
+# suggests) remains the manual path.
+refresh_ellmer_otel_cache <- function() {
+  tryCatch(
+    ellmer:::otel_cache_tracer(),
+    error = function(err) NULL
+  )
+  invisible(NULL)
+}
```
simonpcouch, line 131: Worth investigating + considering upstreaming.

R/tracing.R
```diff
@@ -0,0 +1,228 @@
...
+    {
+      the <- asNamespace("otel")$the
+      if (is.environment(the)) {
+        the$tracer_provider <- NULL
+      }
+    },
+    error = function(err) NULL
+  )
+  invisible(NULL)
+}
```
simonpcouch, line 203: Another bit worth investigating + considering upstreaming.

tests/testthat/test-connect.R
```diff
@@ -0,0 +1,57 @@
...
+    "line3"
+  )
+  calls <- 0
+  local_mocked_bindings(
+    connect_req = function(client, ...) structure(list(), class = "fake_req")
+  )
+  local_mocked_bindings(
+    req_url_query = function(req, ...) req,
+    req_perform = function(req, ...) {
+      calls <<- calls + 1
```
simonpcouch: Not big on `<<-` here.

R/connect.R
simonpcouch, line 1: Are these already available in rsconnect, connectcreds, or connectapi?

R/commons.R
```diff
@@ -113,13 +125,17 @@ Commons <- R6::R6Class(
     #' @param data_sources A [data_source()], or a named list of them.
     #' @param context_layer An optional [context_layer()].
     #' @param semantic_layer An optional [semantic_layer()].
-    #' @param log Where to log conversation trajectories. See [commons()].
+    #' @param log Whether to capture conversation trajectories with
+    #'   OpenTelemetry. See [commons()].
+    #' @param share_with Connect usernames granted access to this content's
+    #'   trajectories. See [commons()].
```
simonpcouch: It's strange to me that these docs live on both `Commons` and its `$new()`.

R/connect.R
simonpcouch: > - connectapi covers the permissions half (`content_add_user()`,
  `get_content_permissions()`, `user_guid_from_username()`) but has nothing
  for `/v1/content/:guid/traces`, which is the client's main job — the
  endpoint is new and still `@visibility internal`.

simonpcouch:
Successfully reproduced on Connect!

PR #38, opened 2026-07-21
Migrate to shinychat's non-module chat API

cpsievert:
## Why

shinychat just deprecated its module-based chat API in favor of a simpler,
non-module one: [posit-dev/shinychat#264](https://github.com/posit-dev/shinychat/pull/264).

For background: `chat_mod_ui()`/`chat_mod_server()` required wrapping your
chat in a full Shiny module (`moduleServer()`), which turned out to be more
structure than most apps need — and it obscured the chat's real DOM id behind
private module namespacing, which made things like slash-command JS events
harder to wire up correctly. The replacement, `chat_ui()`/`chat_server()`,
does the same job (streaming, cancellation, bookmarking/history) but runs
directly in the caller's own session — no module required, and no more
hidden id namespacing. It also gets shinychat's new automatic multi-conversation
history for free, which replaces the old manual `bookmark_on_input`/
`bookmark_on_response` flags.

The old functions are only *soft*-deprecated for now (they still work, with a
warning), but there's no reason for us to carry the extra module indirection
now that a simpler, better-supported path exists, and this package hasn't
shipped externally yet — so there's no cost to switching before anyone
depends on the old API.

## What changed

- `commons_mod_ui()`/`commons_mod_server()` → `commons_ui()`/`commons_server()`,
  mirroring shinychat's rename.
- `commons_server()` no longer takes `bookmark_on_input`/`bookmark_on_response`
  — shinychat's `chat_server()` now persists history automatically.
- Fixed a bug this migration would otherwise have introduced: the
  provenance-pill code used to rely on the module's `session$ns()` to
  correctly target the chat's DOM element. Without the module wrapper,
  `session$ns()` no longer does that, so pills were silently failing to
  attach for any chat id other than the literal string `"chat"`. Fixed to
  resolve the id the same way shinychat's own internals do, and added a
  regression test.
- Regenerated docs (`devtools::document()`) so the man pages match the new
  signatures.

## Test plan

- [x] `testthat` suite passes (one pre-existing, unrelated failure in
      `test-citations.R` exists on `main` too)
- [x] Added a test verifying the provenance pill targets the chat's actual id


simonpcouch (approved):
Thank you!

PR #42, opened 2026-07-22
feat: instrument commons setup paths with OpenTelemetry spans

gadenbuie:
This PR adds additional OTEL tracing throughout key actions in commons, with an eye toward adding tracing around time-consuming actions. The goal is to increase traceability to be able to point definitively at (or rule out) areas of commons that need performance tuning.

## Summary

Adds OpenTelemetry tracing across commons' setup paths (not just the Shiny app use case): data source construction (`data_source()`, including connection, board, and in-process DuckDB frame loading), agent construction (`commons()`/`Commons$new()`), context layer indexing (`context_store()` build vs. `prewarm()` cache-hit/miss), and Shiny server startup (`commons_server()`). Spans use a `commons_<snake_case>` naming convention with `commons.<area>.<field>` attributes, are emitted via a new `local_commons_span()` helper in `R/tracing.R` that no-ops when the `otel` package isn't installed, and aren't gated behind the existing `log` (trajectory-logging) flag since they cover setup, not conversation content. `is_installed()` is now statically imported from `staticimports` instead of `rlang`, removing a namespace conflict.

Two issues surfaced by review and fixed before this PR: `commons_agent_create`'s span used to be created before trajectory tracing was configured in `Commons$initialize()`, so it silently bound to a cached no-op tracer and never exported even with `log = TRUE`; the span creation was moved to after tracing setup. Board-backed data sources (`data_source(board, ...)`) used to recursively call the public `data_source()`, producing a duplicate nested `commons_data_source_create` span; frame-loading is now factored into a shared internal `data_source_frames()` helper used by both the top-level frames path and the board path.

## Verification

```r
withr::local_envvar(COMMONS_TRACES_DIR = tempdir())
agent <- commons(
  ellmer::chat_anthropic(),
  data_sources = data_source(sales = data.frame(id = 1:2, revenue = c(100, 200))),
  log = TRUE
)
recorded <- otelsdk::with_otel_record(agent$prewarm())
vapply(recorded$traces, `[[`, character(1), "name")
# shows commons_agent_create, commons_data_source_create, etc.
```


R/tracing.R
```diff
@@ -1,3 +1,6 @@
+# @staticimports pkg:staticimports
+#   is_installed
```
gadenbuie, line 2: I pulled in `is_installed()` from static imports because it's much much faster than `rlang::is_installed()` and makes it a lot safer to use with the high-level of frequency we want out of `local_commons_span()`

R/tracing.R
```diff
@@ -74,6 +77,33 @@ check_log <- function(log, call = rlang::caller_env()) {
   invisible(NULL)
 }
 
+# Start and activate a span for the calling frame's lifetime, ending when it
+# exits. Unlike trajectory logging, these setup spans aren't gated behind
+# `log = TRUE`: they cover product setup (data source and agent construction),
+# not conversation content, and otel's default tracer provider is a no-op
+# until an exporter is configured, so this is cheap even when tracing is off.
```
gadenbuie, line 84: I decided not to use `log = TRUE` as the signal; with the OTEL tracing you may or may not want to drop `log` entirely.

But the reasoning for decoupling this from `log` is that I'm seeing the commons tracing as an internal dev-facing feature, like tracing in shiny itself. Rather than being about collecting conversation traces directly for evaluation later, the signal here is just "is OTEL tracing turned on?".

R/tracing.R
simonpcouch: Nice, I hadn't heard of this before. I dig it!

simonpcouch (approved):
Looks good! After poking at this, I agree that the delineation from `log` is reasonable.

Issue #40, opened 2026-07-22
`data-dict.yaml` as an interoperable semantic layer format


simonpcouch:
Spent some time this afternoon prodding at the commons/querychat/canvas interoperability story, where data-dict.yaml becomes a standard format that any of the tools could read from.

* @hadley is open to the semantic layer being defined / documented in `data-dict.yaml`.
* The major vendors support grabbing the relevant stuff with a query.
     - Snowflake is `SYSTEM$READ_YAML_FROM_SEMANTIC_VIEW('db.schema.view')` returns the entire definition as YAML in one SQL call
     - Databricks is `DESCRIBE TABLE EXTENDED catalog.schema.my_metric_view AS JSON` gives the Unity Catalog basics, with additional Genie-specific commands later
     - dbt is `SELECT * FROM {{ semantic_layer.metrics() }}`, also yaml
* The major vendors (and Apache Ossie, a spec adopted by the above two plus dbt) need a few additions to the current spec:
    - `metrics` (per-table): named aggregate SQL expressions with a description, e.g. `expr: SUM(order_total * (1 - discount))`. Snowflake calls these `METRICS` in a semantic view; Databricks calls them `measures` in a metric view.
    - `queries`: vetted SQL statements with optional typed parameters (parameter descriptors can reuse the existing column grammar: `type`, `values`, `range`). Snowflake calls these verified queries (`AI_VERIFIED_QUERIES`); Databricks Genie calls them trusted assets (parameterized example queries and SQL table functions).
    - `filters`: named boolean expressions defining reusable segments, e.g. `expr: region IN ('US', 'CA', 'MX')`. Snowflake marks these with `LABELS = (FILTER)` on a dimension (named `filters` in the legacy YAML); Databricks has a metric-view-level `filter` and "filter" SQL expressions in Genie.

So, something like:

```yaml
$version: 0.2.0
name: sales
tables:
  - name: orders
    description: One row per completed order since 2020.
    source:
      sql: sales.orders          # already planned for the spec
    columns:
      - name: order_total
        type: number(quantity)
        units: USD
        range: [0, .inf]
      - name: region
        type: enum
        values: [AMER, EMEA, APAC]
      - name: status_cd
        type: enum
        values: {90: realized, 10: pending, 20: cancelled}
    metrics:                               # NEW: governed aggregations
      - name: net_revenue
        label: Net revenue
        description: Realized revenue net of discounts. Only status 90 counts.
[...]

PR #43, opened 2026-07-24
Implement lazy loading for pins

### LINKED ISSUE #25: Agent construction is very slow when using `pins_board()` with substantial number of pins
It takes ~50 seconds for the tiles agent to start up if I connect it to all the pins (17) for the tiles agent. Previously, the board wasn't a data source and pins were only accessed through measures so this wasn't a problem.

All pins are read in at startup. `data_source_board()` calls `pins::pin_read()` for every pin listed at construction time, and each `pin_read()` has to round-trip to Connect, load the data into memory, and then load it into DuckDB. 

Is some kind of lazy-loading solution a good option? Commons could store the board reference and pin names at construction time, but defer `pin_read()` until that table is needed.

skaltman:
Closes #25. 

Previously, if you used a substantial number of pins as a data source, apps were slow to load because all pins were read into DuckDB tables at startup. 

With this PR, pins are read in lazily:

* `data_source()` with a specified `board` makes a single `pins::pin_list()` call to validate the pin names and returns without reading anything. A pin specified in `data_source()` that isn't in the returned listing triggers a `pins::pin_exists()` check before it's rejected (Connect caps `pins::pin_list()` at 1000 pins). 
* A prewarm feature that downloads as many pins as it can. `commons_server()` starts it during idle time right after startup. The downloads run in a background R process (`callr::r_bg()`) that calls `pins::pin_download()` for each unread pin, so prewarming can't block the main R process. Prewarming only fills the local pins cache. The tables don't get put into DuckDB until they are required (see next bullet), but those reads just parse the cached file and are fast.
*  A pin is read lazily the first time something requires it: a SQL query, the `describe_table` tool, or a measure.
   * For SQL queries, it works by 1) running the SQL the agent requested and 2) if DuckDB returns an error of the type "Table with name [X] does not exist", reading in that pin and retrying. This introduces some latency, but is pretty fast in practice, and is less messy and more correct than name-matching the tables in the SQL. 
   * `describe_table` loads just the one table it inspects. 
   * _Limitation:_ For measures, running a measure loads all of the source's pins, because the tables it queries can't be reliably or easily extracted from the R code. In practice, the hope is that prewarming has already downloaded necessary pins. One reason to keep it this way for now: if measures eventually live in `data-dict.yaml` files with explicit ties to specific tables, we can use those relationships to lazily load just the required pins.

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -54,10 +54,11 @@ commons_server <- function(id, client, ...) {
     attributes = list("commons.server.id" = id)
   )
 
-  # Build the context index during post-startup idle time (while the user
-  # reads the welcome message) rather than inside the first search. Errors are
-  # swallowed: the first search retries the build and surfaces the failure to
-  # the model.
+  # Build the context index and start the background pin-cache download
+  # during post-startup idle time (while the user reads the welcome message).
+  # Errors are swallowed: the first search retries the index build and
+  # surfaces the failure to the model, and an unwarmed pin is simply
+  # downloaded at its first use.
   later::later(function() {
     tryCatch(client$prewarm(), error = function(err) NULL)
   })
--- a/R/commons.R
+++ b/R/commons.R
@@ -289,8 +289,9 @@ Commons <- R6::R6Class(
         )
         context_store(layer)
       }
-      # One failing pin shouldn't stop the others from loading; it stays
-      # pending and is retried on demand.
+      # A pin that fails to read leaves the rest of its source pending, to be
+      # read on demand; wrap each source so one bad source can't stop the
+      # others from prewarming.
       for (source in private$sources) {
         tryCatch(source_ensure_all(source), error = function(err) NULL)
       }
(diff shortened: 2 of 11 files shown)


skaltman:
For the tiles app, this brings the pins source construction down from upwards of 20 seconds to ~4 seconds.

simonpcouch (approved):
Let's do this!

The  piece of this that I do feel unsure about is that because a measure's tables can't be extracted from its R code, running a measure that injects a board source loads *all* of that source's pins. `call_measure` is ultimately the happy path, so I think we'd hit this higher-latency path in the highest-trust cases (which is hopefully most of them). No measures in that tiles agent PR actually do this, because dscoetools is the one reading the pins and they're not defined in the measure formals. I think we can do better here in the future with an explicit tables declaration, either in the measure definitions (e.g. an `@tables` tag) or in the data dictionaries once metrics live there per-table.

Since tiles-agent will not be impacted by this, and this will get much easier to do correctly once we adopt data-dict (#40), we'll defer that for now.

Issue #6, opened 2026-06-24
Broaden semantic layer definition


simonpcouch:
Currently just a set of measures. The usual usage of this term is broader; some of the information that we're currently recommending to push to the context layer should live in the semantic layer itself.

PR #48, opened 2026-07-24
Widen the semantic layer to dictionary-authored definitions

simonpcouch:
Pretty experimental. Want to see how this feels with devrel-io. Related to #40 and #6.

Widens the implementation of commons' semantic layer to accommodate the devrel-io case. This PR tries to anticipate where Hadley is heading in `data-dict.yaml` and prematurely implement support for it, as that shape makes more sense for this use case more so than commons' existing `semantic_layer()` does.

The new elements of the semantic layer can only be added to the harness via `data-dict.yaml`  for now and map more readily to the usual semantic layer elements from other vendors. 

The high-trust path now includes one more tool, `query_metrics`, a normal-semantic-layer-shaped analogue to `call_measure`. `search_measures` becomes `search_pool` and searches across the whole pool of vetted computations, both measures and also definition types: `metrics` × `dimensions` × `filters`, each of which can be composed in `query_metrics`. 

Those definitions can also be substituted in `run_sql`.

When there are no measures (like in the devrel-io case) or no governed definitons (like, for now, tiles-agent), the tools and prompting for that part of the agent's harness is removed.


inst/prompts/system-prompt.md
```diff
@@ -5,25 +5,15 @@ Do not announce tool calls; before your final response to the user, you should o
 
 # How to answer
 
-Registered measures are the preferred way to answer data questions. For any
-question that needs data, your first tool call must be `search_measures` with
-the user's question. Do this even if a table looks easy to query directly. If
-`search_measures` returns a relevant measure, call `call_measure` with the exact
-measure name and argument names returned by `search_measures`.
+Search context with `search_context`, inspect relevant tables with
+`describe_table`, then run a read-only query with `run_sql`.
```
simonpcouch: The preferred path is now `search_pool()` -> trusted computation, whether that's `call_measure()` or `query_metric()`. This seems to deprecate that happy path in favor of `search_context()` -> `run_sql()`, which is not what we want.

R/citations.R
```diff
@@ -145,12 +146,33 @@ add_citation_request <- function(result, tracker) {
...
-citation_request_text <- function() {
-  paste(
-    readLines(
-      system.file("prompts/citation-request.md", package = "commons"),
-      warn = FALSE
-    ),
-    collapse = "\n"
+# Composed from the agent's actual trust surface, so the request never
+# implies operations the agent doesn't have (a measure-less agent's answers
+# are all fallback; there is no "measure alone" path to contrast with).
```
simonpcouch: Lots of these new comments seem to use internal shorthand. Remove every new code comment and then carefully consider what's actually not self-evident from the source code.

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+# fallback from the governed path --- it's the vehicle for it.
+run_sql_description <- function(definitions, measures = list()) {
+  parts <- c(
+    "Run a read-only SELECT query against a data source.",
+    if (length(measures) > 0) {
+      "Use this when no registered measure answers the question."
+    },
+    if (!is.null(definitions) && length(registry_records(definitions)) > 0) {
+      paste0(
+        "Governed definitions (see the system prompt) can be written as ",
```
simonpcouch:
```suggestion
        "Governed definitions can be written as ",
```

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+    character(1)
+  )
+  paste0("Applied governed definitions:\n\n", paste(lines, collapse = "\n"))
+}
+
+# Composed from what the agent actually has: the measure-fallback framing
+# only when measures exist, the token surface only when a dictionary
+# declares definitions. For a definitions-only agent, run_sql isn't a
+# fallback from the governed path --- it's the vehicle for it.
+run_sql_description <- function(definitions, measures = list()) {
```
simonpcouch: Feels like a bit of a smell that we're defining this here rather than in `run_sql()`.

R/pool.R
```diff
@@ -0,0 +1,387 @@
...
+      function(record) {
+        sprintf(
+          "(%s) AS %s",
+          record$expanded,
+          DBI::dbQuoteIdentifier(con, record$name)
+        )
+      },
+      character(1)
+    )
+  )
```
simonpcouch, line 57: Whole lot of `*apply()` over anonymous functions here. Consider whether this could be refactored / substantially trimmed down to read more clearly.

R/pool.R
```diff
@@ -0,0 +1,387 @@
...
+
+# ---- query_metrics -----------------------------------------------------------
+
+# The compiled query surface over dictionary metrics, in the shape semantic
+# layers converge on (MetricFlow's query(metrics, group_by, where),
+# Snowflake's SEMANTIC_VIEW): one tool over the pool. Metric, dimension, and
+# filter names are passed as strings and validated at call time against the
+# registry --- the call_measure convention --- so nothing depends on lazy
+# board sources being probed before tool registration.
+query_metrics_impl <- function(
```
simonpcouch: The happy path is currently `search_pool()` -> `call_measure()`/`query_metrics()`. `query_metrics()` doesn't feel very evocative to me. Let's call it `call_metrics()`.

R/pool.R
```diff
@@ -0,0 +1,387 @@
...
+    slicers <- vapply(
+      same_table,
+      function(r) {
+        sprintf("{{%s}} (%s)", r$name, if (is.na(r$role)) "expression" else r$role)
+      },
+      character(1)
+    )
+    if (length(slicers)) {
+      sprintf("Slicers on this table: %s.", paste(slicers, collapse = ", "))
+    }
```
simonpcouch, line 289: A "slicer" here is a filter elsewhere, right? Can we just call it a filter?

inst/prompts/system-prompt.md
simonpcouch: Ah, I see--other prompting is injected later with `measures_prompt_text()` or friends. The way this is currently worded puts these two section in contradiction. I think we should probably remove the "How to Answer" section completely here and then assemble it programmatically later (internally).

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+# fully-resolved SQL, erroring on reference cycles.
+resolve_definition_references <- function(
+  definitions,
+  table,
+  call = rlang::caller_env()
+) {
+  resolving <- character()
+  expanded <- list()
+
+  resolve <- function(name) {
```
simonpcouch: Not into the function-in-function or the `->>`.

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+        i = "This source has no governed definitions."
+      ),
+      call = call
+    )
+  }
+  available <- vapply(
+    records,
+    function(record) sprintf("{{%s}} (%s)", record$name, record$table),
+    character(1)
+  )
```
simonpcouch: This should just be 'normal' cli inline substitution.

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+      )
+      sprintf("- %s: %s", scope, paste(parts, collapse = "; "))
+    },
+    character(1)
+  )
+}
+
+definition_index_item <- function(record) {
+  item <- sprintf("`{{%s}}`", record$name)
+  label <- flatten_inline(record$label %||% "")
```
simonpcouch: This should just be 'normal' cli inline substitution.

R/definitions.R
```diff
@@ -0,0 +1,730 @@
...
+  call = rlang::caller_env()
+) {
+  if (length(definitions) == 0) {
+    return(list())
+  }
+  definitions <- key_by_name(definitions, "definition", call = call)
+
+  for (name in names(definitions)) {
+    if (!grepl("^[A-Za-z_][A-Za-z0-9_]*$", name)) {
+      token_example <- sprintf("{{%s}}", name)
```
simonpcouch: This should just be 'normal' cli inline substitution.

simonpcouch:
Going to go ahead and merge so that the example agents don't need to depend on this branch and so that I can push #55 forward, accepting that we're relying on data dict entries that may not actually land on `main` and may need to revert parts of this.

PR #54, opened 2026-07-27
sandbox the run_r worker on macOS via Seatbelt

### LINKED ISSUE #50: Implement sandboxes for local eval / testing
For now, we've only implemented a sandbox on Linux because that's what the deployed Connect app will run against. For local eval / testing, implement a sandbox for `run_r` with the same restrictions as the Linux sandbox deployed on Connect. The approach doesn't necessarily have to be exactly the same as the Linux approach (which spawns the process regularly and then engages the sandbox with some C code)—it's it's easier to launch the process in a sandbox, just do that. That said, we would prefer not to require root to do so.

At that point, consider Windows as well. It may be much harder there.

simonpcouch:
Closes #50.

> Adds an __APPLE__ branch to c_sandbox_engage() that self-restricts the worker with sandbox_init() and an SBPL profile mirroring the Linux restrictions: reads limited to R and system roots, writes limited to the work dir and parent tempdir (which need read access too, matching Landlock's FS_V1_ALL on rw roots), and no network — (deny network*) covers unix sockets, so DNS is blocked like seccomp's socket() denial. Roots are normalizePath()ed since Seatbelt matches canonical paths (/tmp and /var live under /private). The denial tests now run on macOS and probe real egress to example.com, including via a subprocess.

### DIFF
--- a/R/sandbox.R
+++ b/R/sandbox.R
@@ -2,9 +2,6 @@
 # sandbox itself in worker_init() (run-r.R), calling the C symbol directly;
 # the parent process is never sandboxed.
 
-# Report what the running kernel supports: the Landlock ABI version (-1 when
-# unavailable), whether seccomp filters can be installed, and whether macOS
-# Seatbelt is available.
 sandbox_capabilities <- function() {
   caps <- .Call(c_sandbox_capabilities)
   list(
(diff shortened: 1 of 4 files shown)


R/sandbox.R
```diff
@@ -3,8 +3,13 @@
 # the parent process is never sandboxed.
 
 # Report what the running kernel supports: the Landlock ABI version (-1 when
-# unavailable) and whether seccomp filters can be installed.
+# unavailable), whether seccomp filters can be installed, and whether macOS
+# Seatbelt is available.
```
simonpcouch: This code is self-explanatory. Remove the comment entirely.

PR #61, opened 2026-07-30
add a sandbox fallback for kernels without Landlock

### LINKED ISSUE #59: fallback for `run_r` sandbox on older Linux kernel versions
It seems like some small subset (Fable later guessed 10-15%) of deployments will not have a new-enough Linux kernel to use Landlock. "5.13+ with Landlock enabled, which in terms of Connect's supported OSes means Ubuntu 22.04+, RHEL 9.6+, or RHEL 10 — and rules out RHEL 8 entirely."

We'll need to decide whether to support an unsandboxed `run_r` fallback. Maybe just error by default and allow users to set some argument, option or envvar to disable the sandbox?

<details><summary>Write-up from Fable</summary>
<p>

Here's the picture. Short version: **Connect's version doesn't matter — nothing in Connect gates or blocks the sandbox. The constraint is entirely the kernel: 5.13+ with Landlock enabled, which in terms of Connect's supported OSes means Ubuntu 22.04+, RHEL 9.6+, or RHEL 10 — and rules out RHEL 8 entirely.**

**What the commons sandbox needs from the kernel.** `src/sandbox.c` uses three mechanisms, and only one is demanding:

- Landlock, via raw `landlock_create_ruleset`/`add_rule`/`restrict_self` syscalls, using only the 13 ABI-v1 filesystem rights (`FS_V1_ALL`). That's the floor introduced in kernel 5.13, so any Landlock-enabled kernel suffices — but Landlock must be compiled in *and* active in the boot LSM list, not just "kernel ≥ 5.13".
- seccomp filter mode via `prctl(PR_SET_SECCOMP, ...)` plus `PR_SET_NO_NEW_PRIVS` — both kernel 3.5+, a non-issue on anything Connect runs.
- `setrlimit(RLIMIT_AS)` — ancient.

Note the failure mode is hard, not graceful: on Linux, `worker_init()` always engages the sandbox and `c_sandbox_engage` raises if `landlock_create_ruleset` fails, so on a kernel without Landlock, `run_r` doesn't degrade to unsandboxed — it errors.

**What Connect contributes.** I pulled and searched `posit-dev/connect` (it was cloned at `~/Documents/rrr/connect`): every seccomp/Landlock reference in the tree is in vendored k8s/golang.org libraries. Connect's own content launching goes through `rsandbox` (`job/sandbox/sandbox.go` → `ext/rsandbox`), which is a setuid helper doing run-as-user, mount/user namespaces, and rlimits — it installs no seccomp filter and doesn't set `no_new_privs`, so it can't block the syscalls the worker needs. The Kubernetes launcher (`kubernetes/resource_bases.go`) sets only `runAsUser`/`runAsGroup`/`supplementalGroups` in the pod security context, no `seccompProfile`. So there's no minimum Connect version; any release runs this fine given a suitable kernel.

**Mapping to Connect's supported OSes** (`docs/admin/getting-started/requirements/index.md`):

| OS                           | Kernel      | Landlock                                                                                                              |
| ---------------------------- | ----------- | --------------------------------------------------------------------------------------------------------------------- |
| Ubuntu 22.04 / 24.04 / 26.04 | 5.15 / 6.8+ | Enabled by default since 22.04 GA (ABI 1 on 22.04, ABI 4+ on 24.04)                                                   |
| RHEL 10                      | 6.12        | Enabled                                                                                                               |
| RHEL 9                       | 5.14        | Only from **RHEL 9.6** (`kernel-5.14.0-568.el9`), where Red Hat enabled it by default and backported up to ABI 5      |
| RHEL 8                       | 4.18        | Never — `run_r` will hard-fail here                                                                                   |
| SLES 15 SP6/SP7              | 6.4         | Kernel is new enough, but I couldn't confirm it's in SUSE's default LSM list — worth a `dmesg \| grep landlock` check |

For off-host/container execution, the kernel is the K8s node's, and since Connect doesn't set a `seccompProfile`, pods run seccomp-unconfined unless cluster policy forces `RuntimeDefault` — and even then, current containerd/Docker default profiles allowlist the landlock syscalls. Consistent with what we saw earlier on connect.posit.it (Landlock ABI 7, everything working).

So the one deployment population that genuinely can't use this is RHEL 8 hosts and RHEL 9 ≤ 9.5 hosts under native execution. If that matters, options are a capability check up front (`sandbox_capabilities()` already surfaces the ABI) with a clearer error, or an explicit opt-out — but as written, the "sandbox must engage" stance means those hosts just can't run `run_r`.

Sources: [Is Landlock LSM available in RHEL? (Red Hat)](https://access.redhat.com/solutions/7028727), [Ubuntu bug #1953192 "Enable Landlock by default"](https://bugs.launchpad.net/bugs/1953192), [landlock(7)](https://man7.org/linux/man-pages/man7/landlock.7.html), [containerd default seccomp profile](https://github.com/containerd/containerd/blob/main/contrib/seccomp/seccomp_default.go)
</p>
</details>

simonpcouch:
Closes #59.

### DIFF
--- a/R/run-r.R
+++ b/R/run-r.R
@@ -38,8 +38,6 @@ tool_run_r <- function(private) {
       "summaries (head(), summary()) over large outputs.",
       if (identical(private$worker$network, "none")) {
         "\n- The session has no network access."
-      } else {
-        "\n- The session has full network access."
       },
       "\n- The session can only write to its own temporary directory."
     ),
--- a/R/run-r.R
+++ b/R/run-r.R
@@ -525,7 +525,7 @@ worker_init <- function(
     callr_data[[".__stderr__"]]
   )
   sym <- getNativeSymbolInfo("c_sandbox_engage", PACKAGE = "commons")
-  .Call(
+  sandbox_backend <- .Call(
     sym,
     read_roots,
     write_roots,
@@ -534,6 +534,7 @@ worker_init <- function(
     preserve_fds,
     network
   )
+  options(commons.sandbox_backend = sandbox_backend)
   invisible(TRUE)
 }
 
--- a/R/run-r.R
+++ b/R/run-r.R
@@ -492,13 +492,15 @@ worker_init <- function(parent_tmp, work_dir, dll_path, sandbox_mode = "auto") {
       USE.NAMES = FALSE
     )
   }
-  # Both mechanisms match symlink-free paths: Connect's packrat library is a
+  # The sandboxes match symlink-free paths: Connect's packrat library is a
   # farm of symlinks into a shared cache, and macOS's /tmp and /var live
   # under /private, so grant resolved paths alongside the originals.
   pkg_dirs <- list.dirs(.libPaths(), recursive = FALSE)
   os_roots <- switch(
     sysname,
-    Linux = c("/usr", "/lib", "/lib64", "/etc", "/opt/R"),
+    Linux = c(
+      "/usr", "/bin", "/sbin", "/lib", "/lib64", "/etc", "/opt/R"
+    ),
     Darwin = c(
       "/usr", "/bin", "/sbin", "/System", "/Library",
       "/private/etc", "/private/var/db", "/opt", "/dev"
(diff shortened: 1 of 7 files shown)


R/run-r.R
```diff
@@ -265,6 +269,32 @@ worker_ensure <- function(worker, fn_sources = character()) {
   invisible(worker)
 }
 
+run_r_sandbox_mode <- function(call = rlang::caller_env()) {
+  mode <- getOption("commons.run_r_sandbox", "auto")
+  modes <- c("auto", "landlock", "userns", "seccomp-only")
```
simonpcouch: I think this should be a matter of capability rather than mechanism. Hopefully, eventually, Connect will just have a solution for this, and we won't need the fallback here. Also, a user testing on a mac shouldn't mean the option (or argument) should change.

Instead, the enum might represent whether to enable network and whether to enable fs. Like `"full", "network-only", "fs-only", "none"` or something

simonpcouch: Could also be

  commons(..., sandbox = c("filesystem", "network"))
  commons(..., sandbox = "network")
  commons(..., sandbox = character())

Or

  commons(
    ...,
    allow_network = FALSE,
    allow_host_filesystem = FALSE
  )

Or

  commons(..., allow = "none")
  commons(..., allow = "network-only")
  commons(..., allow = "filesystem-only")
  commons(..., allow = "all")

simonpcouch: I do think it makes more sense to set this as an argument than as an option.

simonpcouch: Another is

  commons(..., sandbox = sandbox())

  sandbox <- function(
    network = c("none", "full"),
    filesystem = c("none", "full")
  ) {
    network <- rlang::arg_match(network)
    filesystem <- rlang::arg_match(filesystem)
    structure(list(network = network, filesystem = filesystem), class =
  "commons_sandbox")
  }

simonpcouch: I guess the other thing I'm thinking about is that the network is probably the only thing that we want to make toggleable. I can't imagine deploying this on Connect and not sandboxing the file system.

R/run-r.R
```diff
@@ -36,8 +36,12 @@ tool_run_r <- function(private) {
       "\n- Do not use this tool to talk to the user; explanations belong in your reply.",
       "\n- Return results implicitly (`x`, not `print(x)`) and prefer brief",
       "summaries (head(), summary()) over large outputs.",
-      "\n- The session has no network access and can only write to its own",
-      "temporary directory."
+      if (identical(private$worker$network, "none")) {
+        "\n- The session has no network access."
+      } else {
+        "\n- The session has full network access."
+      },
```
simonpcouch:
```suggestion
      },
```

The agent would reasonably assume it has network

Issue #60, opened 2026-07-30
investigate issue reading trajectories from deployed app


simonpcouch:
```r
read_trajectories("<tiles-agent-url>")
#> Error in `resp_body_raw()`:
#> ! Can't retrieve empty body.
```

Same with just providing the GUID.

simonpcouch:
With 

```
    if (!httr2::resp_has_body(resp)) {
      break
    }
```

I just get an empty list.

simonpcouch:
Connect shipped a change to the internal env yesterday that changes the shape of the traces API and the way that apps should interact with it. Old traces will no longer be accessible

PR #62, opened 2026-07-30
fall back to per-job traces when the content trace store is empty

simonpcouch:
Should unblock you for the moment, @skaltman, but I'm not convinced that I really have a handle on the issue yet.

Related to #60.


simonpcouch:
Closing in favor of #63.

PR #63, opened 2026-07-31
fix `read_trajectories()`

### LINKED ISSUE #60: investigate issue reading trajectories from deployed app

simonpcouch:
Closes #60. First commit migrates to the new place where the traces will now live, second commit is a hotfix for a dev Connect issue (41838 on Connect).

### DIFF
--- a/tests/testthat/test-connect.R
+++ b/tests/testthat/test-connect.R
@@ -61,129 +61,54 @@ test_that("connect_trace_lines pages until the total is exhausted", {
 })
 
 test_that("connect_trace_lines reads collector and legacy stores", {
-  local_mocked_bindings(
-    connect_req = function(client, ...) structure(list(), class = "fake_req"),
-    connect_job_trace_lines = function(
-      client, guid, from, enough, page_size, call, lines
-    ) {
-      c(lines, "legacy")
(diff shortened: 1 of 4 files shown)


tests/testthat/test-connect.R
```diff
@@ -57,6 +60,137 @@ test_that("connect_trace_lines pages until the total is exhausted", {
...
+  local_mocked_bindings(
+    req_url_query = function(req, ...) req,
+    req_perform = function(req, ...) 1,
+    resp_has_body = function(resp) TRUE,
+    resp_body_string = function(resp) "current",
+    resp_header = function(resp, name) {
+      if (name == "Server") "Posit Connect v2026.07.0" else "1"
+    },
+    .package = "httr2"
+  )
```
simonpcouch: We should drop the test if we need to do this much mocking (or `skip_if_*()` / `skip_on_cran()` and then hit a real endpoint)

simonpcouch: Yeah, same story for all of the tests in this PR.

PR #76, opened 2026-08-03
make prompt template easier to read / reason about

### LINKED ISSUE #67: make prompt template easier to read / reason about
...both for us as authors and for app authors.

simonpcouch:
Closes #67.


inst/biodiversity-app.R
```diff
@@ -84,12 +84,17 @@ greeting <- paste(
...
 
 system_prompt <- paste(
-  ellmer::interpolate_file(
+  readLines(
     system.file("prompts/system-prompt.md", package = "commons"),
-    date = Sys.Date()
+    warn = FALSE
   ),
+  collapse = "\n"
+)
```
simonpcouch: The example app just shouldn't touch the system prompt all. You can remove the `sandbox_backend` stuff.

simonpcouch:
Definitely some tension between the idea that we should let developers edit the system prompt freely (which they will want to) and commons' own ability to iterate on the system prompt (and push those improvements to upstream users automatically).

inst/prompts/system-prompt.md
```diff
@@ -1,10 +1,62 @@
...
+<!-- The governed workflow must agree with the tools registered for this agent. -->
+{[ if (has_governed_operations) "Governed operations are the preferred way to answer data questions:" else "" ]}
+
+{[ if (has_measures) "- Run a measure with `call_measure`." else "" ]}
+{[ if (has_metrics) "- Compute a metric with `call_metrics`." else "" ]}
+{[ if (has_definitions) definition_action else "" ]}
+
+<!-- A searchable pool contains measures or more definitions than fit below. -->
+{[ if (has_governed_operations && has_search_pool) r"(
+For any question that needs data, your first tool call must be `search_pool` with the user's question. Do this even if a table looks easy to query directly, and use the exact names it returns. Do not call `run_sql` or `describe_table` until after you have.
```
simonpcouch:
```suggestion
For any question that needs data, your first tool call should be `search_pool`. Do this even if a table looks easy to query directly, and use the exact names it returns. Do not call `run_sql` or `describe_table` until after you have.
```

R/commons.R
```diff
@@ -30,19 +30,30 @@
...
-#'   [ellmer::interpolate_file()]. commons appends documentation of the
-#'   available tables and data dictionaries to the prompt itself, along with
-#'   a "How to answer" workflow section assembled from whichever of
-#'   registered measures and governed definitions the agent actually has;
-#'   the file needn't (and shouldn't) describe them.
+#'   commons renders conditional sections from the agent's composition and
+#'   interpolates runtime values such as its table roster. Custom templates
+#'   may edit, remove, or reposition any section. Commons expressions open
+#'   with `{[` and close with `]}`, leaving ellmer's `{{ }}` delimiters
+#'   available for your own substitutions:
```
simonpcouch:
```suggestion
#'   commons renders some sections conditionally and
#'   interpolates runtime values such as its table roster. Custom templates
#'   may edit, remove, or reposition any section. Commons expressions open
#'   with `{[` and close with `]}`, leaving ellmer's `{{ }}` delimiters
#'   available for your own substitutions:
```

R/commons.R
```diff
@@ -30,19 +30,30 @@
...
+#'   )
+#'   commons(
+#'     # ...
+#'     system_prompt = as.character(system_prompt)
+#'   )
+#'   ```
+#'
+#'   A `{{organization}}` expression is resolved by ellmer, while commons'
+#'   template expressions remain untouched. Commons expressions are evaluated
+#'   as trusted R code when the agent is created.
```
simonpcouch:
```suggestion
```

simonpcouch:
There should probably be some preferred way for users to append things to the system prompt without actually copying the file (like `always` worked).

There could be some sort of `options = list(prompt_introduction)` etc etc that allows people to customize the prompt without literally copying it. Or maybe this would be the only entry point to changing it.

Issue #79, opened 2026-08-03
Carry trust tags through OTel


skaltman:
Right now, trust tags do not persist in the OTel logs. As a result, the trajectory reviewer has to reconstruct the tags from the conversation (using the tool call names and citations), introducing duplication and the possibility of errors. 

It would be nice if the trust tags were carried through OTel so that they are available when reading in trajectories.

skaltman:
Another candidate for information carried through OTel: the conversation titles from the shinychat history, so the reviewer app doesn't need to reconstruct them.

simonpcouch:
Related to #107—citations and trust tags are now generated as part of the stream. We should confirm that they're persisted in the Connect OTel logs now and then remove the complexity in the app that exists to regenerate them.

PR #80, opened 2026-08-04
Add trajectory reviewer

### LINKED ISSUE #53: Viewing trajectories
A couple thoughts:

* In the app, might be nice to have:
    - A "hit rate" with the rates of the different trust levels across the conversations (maybe with a time period selector? or a +/- across time?)
    - A conversation viewer that puts the CONVERSATIONSUMMARY that shinychat would use in the normal conversation selection (the beginning of the first user message?), the number of users turns, and the UI pills that appeared in the conversation. Then, when clicked, see the commons UI where the "Submit" button from the usual chat actually saves feedback rather than "submitting" in the conversation
* "Trajectory" should probably renamed to "Log" or "Trajectory" s.t. `read_trajectories()` becomes `*_read`, and then the viewer has the same prefix but is `*_view()`

cc @skaltman !

skaltman:
Adds a trajectory reviewer app. 

Launch the reviewer app with:

```r
# launches reviewer for tlg agent trajectories
trajectory_review(read_trajectories("04ee623c-3daf-4893-989e-7be4f2c0e7a7"))
```

Trust tags are reconstructed from the conversation, see https://github.com/posit-dev/commons/issues/79. 

Notes and flags are written to an append-only JSONL file. There is probably later work to do on how the review data is stored. You can join the review data back to the trajectories data for analysis:

```r
trajectories <- read_trajectories("04ee623c-3daf-4893-989e-7be4f2c0e7a7")
trajectory_reviews_read("commons-review.jsonl", trajectories)
```

image

Addresses https://github.com/posit-dev/commons/issues/53


R/trajectory-review-log.R
```diff
@@ -0,0 +1,239 @@
...
+#'   `COMMONS_REVIEW_FILE` environment variable when set, and otherwise to
+#'   `"commons-review.jsonl"`.
+#' @param trajectories An optional named list of conversations returned by
+#'   [read_trajectories()]. When supplied, review records are joined to their
+#'   conversation or exchange.
+#'
+#' @return A list of actionable review records in file order: all notes and
+#'   currently active flags. When `trajectories` is supplied, each record also
+#'   contains `turns`.
+#' @export
```
simonpcouch: Part of me wonders whether we ought to write the reviews with the turns 'joined' in in the first place. Since it's a human that reviews the responses, I would guess that the scale of these wouldn't reach past 100s of comments, so these files wouldn't ultimately end up huge.

The reason I say this is because I would suspect that humans would not be the ones to then go and read in all of the reviews + process them, so I'd prefer we not export a function mostly for the coding agent's usage. Instead, that format that contains both the turns and the reviews could be document with a few-liner in the `trajectory_review()` docs that gets the agent most of the way there.

R/trajectory-timeline.R
```diff
@@ -0,0 +1,410 @@
...
+}
+
+# A 100%-stacked area chart of each bin's trust-level shares -- one stacked
+# column when only one bin is dated, since a single point can't make an
+# area. The tooltip is one card drawn wholly from per-bin text -- unified
+# hover's date header can't carry the bin's n beside the date -- and every
+# trace carries the same card: with closest-point hover, the label then
+# anchors to whichever band boundary sits nearest the pointer instead of
+# always to the chart's top edge. The card header's legend names the
+# levels, so the widget's own legend stays off.
```
simonpcouch: Many of the comments here are quite extensive and document the 'what' more than the 'why'. The code itself should be self-explanatory when possible and code comments should document decisions / architectures that would otherwise be surprising. This is unfortunately just something that the recent Claude models really struggle with.

I might have Sol read in the whole diff of these changes, familiarize with them as needed, then delete all of the newly added comments and add brief ones only where it thinks there are reasonable.

R/trajectory-read.R
```diff
@@ -51,7 +51,11 @@
 #' ```
 #'
 #' @return A list of conversations, named by conversation id and ordered
-#'   oldest-first. Each conversation is a list of [ellmer::Turn]s.
+#'   oldest-first. Each conversation is a list of [ellmer::Turn]s and carries
+#'   a `last_active` attribute: a `POSIXct` giving the time of the
+#'   conversation's most recent chat activity. The list carries a `source`
+#'   attribute identifying the local trace directory or Connect content from
+#'   which it was read.
```
simonpcouch, line 58: I had wondered whether there was already some metadata along these lines inside of the Turn objects, but I don't think there is. Attributes always have a bit of a smell to me, but I don't see an obvious alternative.

inst/www/commons-viewer/commons-viewer.css
```diff
@@ -0,0 +1,418 @@
...
+    to right,
+    transparent calc(50% - 0.5px),
+    var(--bs-border-color, #dee2e6) calc(50% - 0.5px),
+    var(--bs-border-color, #dee2e6) calc(50% + 0.5px),
+    transparent calc(50% + 0.5px)
+  );
+  cursor: col-resize;
+  touch-action: none;
+  width: 0.5rem;
+}
```
simonpcouch: I'm wondering whether this complexity could be cut out in favor of bslib's resizable sidebar, with e.g.

```
  1. Use bslib’s resizable sidebar

  bslib::layout_sidebar(
    shiny::uiOutput("transcript"),
    sidebar = bslib::sidebar(
      shiny::uiOutput("review_bar"),
      position = "right",
      width = 320,
      resizable = TRUE
    )
  )
```

Sol's guess:

>   This could remove the custom pane-resizer HTML, pointer/keyboard JavaScript,
  responsive grid logic, and associated CSS. bslib’s implementation already
  supports keyboard resizing and accessibility.

I might ask GPT (probably not Claude) to cut out all of that complexity in favor of `layout_sidebar` and see what happens.

R/trajectory-review.R
```diff
@@ -0,0 +1,1089 @@
...
+        NULL,
+        value = draft,
+        placeholder = if (whole_conversation) {
+          "Add a note about this conversation"
+        } else {
+          "Add a note about this exchange"
+        },
+        rows = 2,
+        width = "100%"
+      ),
```
simonpcouch: Another simplification opportunity from GPT. This might not be a huge win, but worth a try:

> 2. Use bslib::input_submit_textarea()

```
  bslib::input_submit_textarea(
    "review_note",
    placeholder = "Add a note",
    submit_key = "enter"
  )
```

> Together with update_submit_textarea(), this replaces the custom textarea/
  button, Enter-key JavaScript, and much of the draft synchronization. It would
  work best if the composer remained mounted rather than living inside a
  repeatedly rendered uiOutput().

> Both bslib substitutions require bslib 0.11.0; the textarea also requires
  Shiny 1.11.1.

R/trajectory-review.R
```diff
@@ -0,0 +1,1089 @@
...
+          value = c(dates$min, dates$max),
+          minDate = dates$min,
+          maxDate = dates$max,
+          dateFormat = "MMM d, yyyy",
+          update_on = "close",
+          addon = "none",
+          # With toggling on, clicking the range's start date a second time
+          # deselects it, making a one-day window unreachable.
+          toggleSelected = FALSE
+        ),
```
simonpcouch, line 398: Possibly could be:

```
shiny::dateRangeInput(
   "window",
    "Dates",
    start = dates$min,
    end = dates$max,
    min = dates$min,
    max = dates$max
  )
```

This would allow for removing the `shinyWidgets` dependency. That said, it's fine with me to keep the dependency around; only a Suggests.

simonpcouch (commented):
Super excited about this! Some initial thoughts here.

simonpcouch:
Feedback from Paul & Garrick:

* Trajectory is good, no need to rename
* Trust level might need some interface other than a selector
* The Conversation vs. Question picker is confusing. What does it mean to be in one tab or another?
* Should the timeline plot instead be a series of stacked bars?
* Maybe the conversation view should be full-height and then the plot becomes the top of the sidebar that the "Notes for Question n" currently lives in
* Should there be a way to remove / edit existing notes?

R/trajectory-review-log.R
simonpcouch: Garrick thinks we _should_ probably have a function like this. He also suggested maybe this ends up being a directory of markdown files so that they're super agent-readable (whether that's immediately what they become or get post-processed into after this intermediate state)

R/trajectory-review.R
skaltman: The reason for `shinyWidgets::airDatepickerInput` is so that the date picker can appear as a single calendar and you can select non-continuous ranges. That said, this probably isn't important and we can just use `shiny::dateRangeInput()`.

R/trajectory-review-log.R
skaltman: I like Garrick's idea, and I think regardless of format, the review writing and ingesting process needs a second look. For now:

* I removed `trajectory_reviews_read()` for the reason you stated, and because the trajectory review format may change. It's not needed to read in the trajectories, so this doesn't result in any loss of functionality. 
* I did not change the review `jsonl` format to include the turns because the format may change and I am also a bit hesitant to duplicate that data (but could definitely be convinced that this isn't a problem). 
* I can plan to revisit the review log format for v2. 

How does that sound?

skaltman:
@simonpcouch thank you! 

* Did another pass over the comments to make them less verbose and only present when necessary.
* Removed `trajectory_reviews_read()`: https://github.com/posit-dev/commons/pull/80#discussion_r3717497425
* Simplified the notes panel, note text area input, and date picker using built-in shiny/bslib functions (`bslib::layout_sidebar`, `bslib::input_submit_textarea`, `shiny::dateRangeInput`) 
* The timeline now uses stacked bars for sparse data. Binning dates in a stacked area chart was confusing because an area chart implies a continuous x-axis.
* Another simplification: The trajectory reviewer now [passes saved messages directly to shinychat](https://github.com/posit-dev/commons/pull/80/changes/785742fcc422c65f696ae8670a3fe5a08a30bcd5) when
  creating the chat, instead of creating an empty chat and replaying each message. 

I think we can defer the changes that came out of the feedback from Paul and Garrick to a follow up PR. Specifically, I think it's worth revisiting how review data is stored and possibly using [markdown like Garrick suggested](https://github.com/posit-dev/commons/issues/82).

simonpcouch (approved):
Great, thanks so much! Agreed that we can iterate on the review reading and other UI points after merge.

In addition to merging upstream, I went ahead and did the `read_trajectories()` -> `trajectory_read()` rename, so the function names are consistent, and added you as author.

PR #81, opened 2026-08-04
Add Connect deploy workflows with PR previews

tdstein:
## Summary
- Adds `.github/workflows/deploy.yml`: sends the app to two Connect servers (dev and staging) using [posit-dev/connect-actions](https://github.com/posit-dev/connect-actions). A push to `main` sends the real, live version. A pull request sends a draft preview instead, and posts a comment with the preview link.
- Adds `.github/workflows/cleanup-previews.yml`: deletes the draft previews from both servers when a pull request closes.
- The two server addresses are stored as GitHub secrets (`CONNECT_SERVER_DEV`, `CONNECT_SERVER_STAGING`) instead of being written directly in the workflow file. This keeps them out of a plain GitHub search. It does **not** fully hide them, though: the deploy step still posts the real address in its pull request comment (see below).
- Login uses Trusted Publishing (OIDC), so no API key is stored. Each app needs a trusted publisher set up for this repo on its Access tab in Connect.
- Fixes the `shinychat` remote in `DESCRIPTION`: the R package actually lives in the `pkg-r` folder of that repo. The old remote pointed at the wrong spot, so it quietly used the CRAN version (0.4.0) instead, which is missing `chat_server`.


inst/manifest.json
simonpcouch, line 1: Could we add this to Rbuildignore so that it doesn't ship with the package?

simonpcouch (approved):
Nice, let's do this! A couple things to address before merge:

- What happens if I make PR A, then PR B, then want to look at the preview for PR A? In other words, is the deployed preview per-branch or per-repo?
- When I click '"Back to active version"', am I actually seeing the most recent deployed preview or am I seeing some previous version of the content, pre-deploy-preview for the given PR?
- Could you add Sara, Garrick, and Carson as viewers of these pages as well?

Noting for them that, for now, there seems to be a Connect issue where, after clicking the link, you will see a page that says you don't have access to the listing. Click "Back to active version" and then you'll see the page.

tdstein:
> * What happens if I make PR A, then PR B, then want to look at the preview for PR A? In other words, is the deployed preview per-branch or per-repo?

It's per branch, so each PR get's their own unique build on Connect!

> When I click '"Back to active version"', am I actually seeing the most recent deployed preview or am I seeing some previous version of the content, pre-deploy-preview for the given PR?

Technically, it's the most recent deployed version, but moving forward that will be whatever is on `main` once this change lands.

> Could you add Sara, Garrick, and Carson as viewers of these pages as well?

Done!

PR #86, opened 2026-08-05
database catalog and semantic model interoperability

simonpcouch:
Related to #55. Just to actually see what one e2e take on this could feel like.

Treat a connection to Snowflake/Databricks plus a schema path as minimally sufficient:

```
db <- DBI::dbConnect(odbc::snowflake())

source <- data_source(
  db,
  options = data_source_options(
    include = DBI::Id(...)
  )
)

agent <- commons(chat_anthropic(), source)
```

commons then discovers the schema and native semantic models without requiring a data dictionary.

Issue #55, opened 2026-07-27
Per-viewer access to data and facts


simonpcouch:
commons agents are built by data scientists who have access to different information than the intended users of the app.

For agents scoped to smaller domains, this isn't as much of an issue; only those who have access to the underlying data should be given access to the app.

For more general data agents, the question is trickier. What happens when the CEO looks at the app vs. a product owner multiple levels down?

There are two axes in question here:

* **Row-level access** is more important and should be more straightforward, at least for Snowflake and Databricks. Run every query as the viewer via connect's viewer-based credentials. Document `data_source(DBI::dbConnect(odbc::snowflake()), ...)` inside `server()` as the happy path (explaining that connectcreds/odbc will do the heavy lifting here). Elsewhere—Postgres, MySQL, SQL Server—the story is trickier. pins would be per-pin via a visitor API key and need per-session board auth (today the board source downloads is process-wide).
* **Fact-level access** is less important and more tricky. In `data_source(con, data_dictionary)`, the machine-readability of data dictionaries should be helpful for us in that only the tables from `con` that the user has access to should be searchable by the model.`context_layer()` is harder. 

There should probably be GRC-focused vignette that, among other things, makes it clear we're not putting any trust in the LLM itself to not leak information.

<details>
<summary>Initial prompt to Fable</summary>

> Up to this point, I've largely tabled the GRC question with commons. Notably, commons agents are built by data scientists who might have access to different information than the users of the app. To some extent, this can be solved by only allowing access to the application to certain people inside of Connect; this works already. That said, We imagine that some folks will take this agent and make it relatively general, like a "Data at \[company\] agent". This means that the privacy story might be rather tricky, e.g. what happens when the CEO looks at the app vs. a product owner 3 levels down?
> 
> There are at least two big questions here:
> 
> * Row-level access permissions on the data itself. For data that's inside of Snowflake or Databricks or whatever, I think this should actually be relatively easy? We probably don't need a story for data that is inside of playing tables, for example.
> * Fact-level access permissions. e.g. should anyone with access to the app have access to any of the information in the data dictionary or context layer?
> 
> Via Connect, we have access to some bits of information about who it is that is looking at the app at a given moment. You have access to the Connect source code--feel free to clone it into inst/ and read the relevant pieces. You also have access to a Connect API, so feel free to experiment against the live service. https://github.com/posit-dev/connectcreds might also be useful.
> 
[...]

PR #96, opened 2026-08-10
add live warehouse test helpers

simonpcouch:


simonpcouch:
PR 1 of many that will give another go at #86 in much smaller pieces. Related to #55.

PR #97, opened 2026-08-10
database source options

simonpcouch:
Related to #55, another smaller portion of #86.


simonpcouch:
This isn't needed—we'll just use `tables`.

PR #103, opened 2026-08-11
Iterate on hex sticker

skaltman:
The image without the stream serves as the reference image for the one with the stream -- only because I liked how the bird looked in that one and couldn't reproduce it again.


simonpcouch (approved):
Nice!!!

I shuffled the .yaml files around so that the original one that generated the one that the second uses as a reference is still around.

I also adjusted the crop so that the bird and the commons plaque are in the center of the sticker. That _does_ mean that the tree is clipped quite heavily and only the slightest bit of the best is visible in the image. What do you think?

If you'd like the tree to be more legible (or removed entirely), we can tear it all down and start with a new bananarama.yaml that integrates the changes from `-stream.yaml` and one-shots the image. Then, we could slot the image that we're ultimately happy with into the same `make_hex.py`.

skaltman:
This looks good to me! Thanks for indulging my tinkering 😆

Issue #106, opened 2026-08-12
Audit trail silently drops conversations/branches: conversation_id never rotates across shinychat history operations


cpsievert:
## Problem

`Commons` mints `private$conversation_id` exactly once, in `initialize()`, and nothing ever rotates it. But shinychat's history controller reuses **one client object for every logical conversation in a session**:

- **New chat** (`HistoryController$new_chat`): `client$set_turns(list())` — same client.
- **Conversation switching** (`switch_to`): `set_turns()` to a completely different stored conversation — same client.
- **Edit** (`handle_edit`): truncates turns to the fork parent via `set_turns()`, then resubmits.
- **Branch navigation** (`handle_navigate`): `set_turns()` to a different leaf of the message tree.

So in the trace, every conversation a user touches in one Shiny session is recorded under a single `commons.conversation.id`.

## Impact: exchanges silently vanish from the audit trail

`build_trajectories()` keeps only the **latest chat span per conversation id** as the canonical history. Two consequences:

1. **New chat / switch:** a user has a 5-exchange conversation, clicks "New chat", asks one question. The latest chat span under the shared id carries only the new conversation's turns — the first conversation's five exchanges disappear from `trajectory_read()` entirely.
2. **Edits / branch navigation:** after an edit-and-regenerate (or navigating back to an earlier sibling and continuing), only the final active path survives review. Answers the user actually saw — rendered with a trust verdict — become unreviewable.

For a governance/audit tool this is the worst kind of gap: it's silent. The reviewer has no signal that anything is missing.

Note the exchange-provenance matching added in the server-side citation verification work already anticipates the id-sharing (see "switched conversations do not donate audit records" and "edited paths retain shared-prefix records and drop abandoned records" in `test-trajectories.R`) — records are correctly never cross-contaminated between logical conversations. That defends the *records*; the *turn content* still evaporates via latest-span-wins.

## Recommended fix

**Rotate `conversation_id` whenever history diverges.** Have `Commons` override `set_turns()` (or detect at `stream_async()` time) that the current turns are not an extension of the last-streamed state, and mint a new id.

Because every chat span carries its full history prefix, the new id's spans still carry the restored/forked history — nothing is lost, and `latest_chat_spans()` stops collapsing distinct conversations.

Trade-off: one logical conversation restored across sessions appears under multiple ids, so review shows some duplicated prefix history. Duplication beats disappearance for an audit tool, and the prefix-comparison machinery in `trajectory-read.R` could dedupe later.

[...]

PR #107, opened 2026-08-13
Verify citations in-stream and record answer provenance for trajectory review

cpsievert:
## Summary

This PR reworks how a commons answer's trustworthiness travels from the model to the user, and
from the user's session to the audit trail. The verification logic itself isn't new — quotes
were already exact-matched against the corpus on the server. What changes is the architecture
around it, in two ways:

1. **What the user sees is now authored by the server.** Citation markup is verified and
   rewritten *while the answer streams*, so only server-authored elements ever reach the
   browser. The client-side retrofit protocol is gone.
2. **What the reviewer sees is now recorded, not reconstructed.** Each answer's trust verdict
   and citation decisions are written to the trace at the moment they're made, and
   `trajectory_review()` plays them back verbatim instead of re-deriving them heuristically.

Together with a conversation-identity fix (#106), the end state is: the pill a user saw and
the pill a reviewer sees are provably the same thing, and every conversation a session touched
is actually present in the audit trail.

## The live path: stream → scanner → asides

Today, the model's raw `<citation>` markup streams into the browser DOM, and after the turn
completes, the server sends verdicts over a custom message so `commons-chat.js` can replace
the rendered elements positionally. That final hop is fail-open: until the JS runs (or if it
miscounts), the user is looking at unverified markup dressed as citations.

Now `Commons$stream_async()` wraps every answer in a streaming scanner (`R/citation-scan.R`):

- The model writes a new block dialect, `<commons-citation>` — a short explanation plus a
  blockquote carrying the verbatim excerpt (`inst/prompts/citation-request.md`). The block
  form replaced the old single-line attribute form because it survives streaming: the scanner
  can hold back a partially-arrived block and release everything else immediately.
- Each quote is verified against the same corpus as before (context docs, measure schemas,
  dictionary entries), but mid-stream. Verified citations are rewritten inline as
  `<shiny-aside>` elements naming their source; failed ones are dropped.
- This is fail-closed and unconditional: the scanner runs whether or not tracing is on, the
  dialect never reaches the browser unprojected, and model-spoofed `<shiny-aside>` markup is
  neutralized. Stored turns keep the raw model text — projection is a display concern.
- After the stream, the answer's tag is derived (`derive_provenance_tag()` in
  `R/provenance.R`): governed-tool answers are "A", fallback-tool answers with a verified
  citation "B", without one "C" — same semantics as before, with "B beats A" because the
  citation is the thing the user is being asked to trust. A/C answers get a compact
  provenance aside appended to the stream; a B answer's provenance UI is its citation asides.

Display rides shinychat's aside component, so the custom pill/citation protocol is deleted:
`commons-chat.js` shrinks by ~225 lines, `commons_server()` sends no custom messages, and
`R/tagging.R` (the old post-hoc provenance derivation) is gone entirely.

## The recorded path: spans → trajectory read-back → review

Previously the trajectory viewer *inferred* trust after the fact — tool names implied the tag
(`call_measure` ⇒ verified, `run_sql` ⇒ cited-or-untrusted) and citations were re-scraped from
answer text, with footnotes explicitly labeled "unverified". An inference can disagree with
what the user was actually shown, which is the wrong failure mode for an audit surface.

Now the decisions are recorded where they're made: `stream_async()` writes the tag
(`commons.provenance.tag`) and the scanner's accept/reject decisions
(`commons.citation.candidates`, JSON) onto the `commons_conversation_turn` span.

On the way back out, `trajectory_read()` (`R/trajectory-read.R`) reads those attributes off
every recorded call in a conversation and attaches them per-exchange to the reconstructed
turns. The one genuinely subtle design decision is *how* a recorded call is matched to an
exchange: structurally, by comparing canonicalized exchange-content signatures as a prefix of
the final history — not by ordinal position. Conversation histories aren't append-only once
shinychat's edit and branch-navigation features are in play, so a position recorded by one
call can name the wrong exchange in another call's reality. The comment above
`exchange_signature()` and the provenance block of `test-trajectories.R` (linear, edited,
restored, switched, conflicting, incomplete) are the spec for this; I'd start a review of the
read-back side there.

`trajectory_review()` then becomes a playback device: it reconstructs citation asides from the
stored raw turns using the same scanner, resolved by the *recorded* decisions rather than a
corpus (it never re-verifies), surfaces rejected-citation counts in a review-audit aside, and
takes its pill copy from the same display table the live aside uses so the two can't drift.
Missing or conflicting records fail closed to "no claim" — the viewer never invents a verdict.

## Conversation identity (#106)

While wiring the recorded path I found that shinychat's history feature reuses one client
object across new chats, conversation switches, and edits (all via `set_turns()`), while
commons minted one conversation id per client — so every conversation a session touched shared
one id, and `trajectory_read()`'s latest-span-wins reconstruction silently kept only the last
one. Fixed in two layers (one commit, `R/commons.R` + `R/chat.R`):

- `stream_async()` rotates the conversation id whenever the current turns no longer extend
  the last streamed state.
- `commons_server()` persists the id in each saved conversation's app-state values via
  shinychat's `on_save`/`on_restore` hooks, so reopening a conversation continues under its
  original id instead of duplicating its history under a fresh one. This layer is commented
  as deletable once shinychat records stable conversation identity on its own spans.

## Reviewing guide

- Scanner semantics (hold-back, projection, spoof handling): `R/citation-scan.R`, specced
  exhaustively in `test-citation-scan.R`.
- Recording: `stream_async()` in `R/commons.R`; tag semantics and display table in
  `R/provenance.R`.
- Read-back and structural matching: `R/trajectory-read.R` + the provenance tests in
  `test-trajectories.R`.
- `R/trajectory-review.R` is mostly deletion of the inference path plus playback.
- Real-browser coverage (streaming apps driven via shinytest2/chromote):
  `test-citation-browser.R` and `tests/testthat/apps/` — hence the new Suggests.

## Trade-offs and notes

- **Old traces go dark.** Pre-existing conversations carry no recorded provenance, so they
  show no trust pill in the reviewer. Deliberate: no claim beats an inferred claim, and the
  package's pre-release stance means no compat shims.
- **Read cost grew.** `trajectory_read()` now parses one chat span per recorded exchange
  rather than one per conversation. Fine at review scale; noted in case it ever matters.
- **Abandoned edit branches** are preserved in traces (each under its own id after #106) but
  the reviewer doesn't yet present them as siblings of the edited conversation — tracked as
  the open half of #106.
- Requires shinychat > 0.4.0 for the aside component; adds an R-CMD-check workflow.


R/citations.R
```diff
@@ -2,118 +2,166 @@
...
+  icon <- citation_icon_url(kind)
+  reason <- if (nzchar(explanation)) paste0("**", explanation, "**\n\n") else ""
+  sprintf(
+    '<shiny-aside label="%s"%s>%s> %s\n\n*Quoted verbatim; matched exactly.*</shiny-aside>',
+    escape_attr(label),
+    if (is.null(icon)) "" else sprintf(' icon="%s"', escape_attr(icon)),
+    reason,
+    trimws(quote)
+  )
 }
```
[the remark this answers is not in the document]
cpsievert: Fixed in a90bc5f: multiline evidence now prefixes every line with Markdown blockquote syntax, with a regression test.

simonpcouch:
Whole lot of red due to GH outages, but going to go ahead and merge!

I made some changes related to the appearance of the pills/citations: the text in the pills themselves is once again unbolded, the citations are shown as superscript footnotes rather than pills, the model's explanatory text is unbolded, and the "Quote verbatim" text is removed, and the "Untrusted" and "Verified answer" text isn't duplicated on hover.

simonpcouch:
Thanks so much for making this happen.

inst/www/commons-chat/commons-chat.css
```diff
@@ -304,6 +304,56 @@ shiny-chat-container .shiny-aside-pill img {
...
+}
+
+shiny-chat-container
+  .shiny-aside-pill:has(img[src*="/citation-"])
+  img,
+shiny-chat-container
+  .shiny-aside-pill:has(img[src*="/citation-"])
+  .shiny-aside-pill__label {
+  display: none;
+}
```
cpsievert, line 355: @simonpcouch Does it feel like any of this should get moved over to shiny chat? If so, want to give me a high-level description of what the aim is here?

inst/prompts/citation-request.md
```diff
@@ -14,11 +14,14 @@ A brief explanation of how the quote supports the answer.
 
 Rules:
 
-- Start `<commons-citation>` at the beginning of a line, exactly as written,
-  with no attributes.
+- Put the first `<commons-citation>` on the line immediately after the answer,
+  with no blank line before it. Start the tag at the beginning of the line,
+  exactly as written, with no attributes.
```
cpsievert, line 19: For my own understanding, what's the reasoning for this approach? Is the concern that if citations are inline in the answer, folks are going to miss them?

inst/www/commons-chat/commons-chat.css
simonpcouch: It might be mostly personal preference. These changes remove the citation (as, not "Verified answer" or "Untrusted", but a citation to a previous tool result) pills in favor of a superscript, which feels like a more neutral and more minimal way of expressing the response is neither necessarily trusted or untrusted and that there's more information that the user can look at if they want to.

inst/prompts/citation-request.md
simonpcouch: This prompting is actually a bit clunky given what the UI will actually do to the citations. I want the citations to appear immediately after the response but models don't seem to want to inline XML on the same line, so I just ask to provide it immediately after and then "lift" it back in line with the previous line.

inst/www/commons-chat/commons-chat.css
cpsievert: One reason I like the pills is that we can give a nice visual clue as to what type of information the citation is referring to through both an icon and a label. 

That's fine if you prefer superscripts, but this doesn't feel like the cleanest way to move in that direction. For one, we wouldn't need the image icons that this CSS is referring to now. More generally, though, I'd be open to officially supporting this as a rendering mode in shinychat.

cpsievert: Looking at this a bit more closely, I think this change also introduces some correctness issues, due to how de-duplicating works in `<shiny-aside>` that have the same labels. Here is a relevant excerpt from Claude:

```
### Numbered markers don't survive more than one citation

This is the one I'd push back on hardest, because the new prompt makes it the common case. shinychat merges adjacent asides that carry a non-empty label into a single shiny-aside-group (jI()/kS() in shinychat.js). I confirmed in a real browser with two attached citation asides:

2 asides, different sources → 1 pill:
  aria-label: "documentation (+1 more)"
  ::before:   counter(commons-citation)     → renders "1"
  .shiny-aside-pill__overflow: display:block, text "+1"

2 asides, same source → 1 pill, visible text just "1"

So:
- The counter never reaches 2. "Restore numbered citation markers" yields a constant 1.
- The second source's name never appears on the pill face at all.
- When sources differ, the marker visually reads "1+1" — the CSS hides img and .shiny-aside-pill__label but not .shiny-aside-pill__overflow, which computes to display: block inside an inline-flex superscript.

The existing browser tests don't catch this: the fixture has exactly one verified citation, and the new R-level test ("adjacent verified citations attach to the preceding text") asserts on the R output string only, never on the rendered DOM.

There's a cleaner path available: shinychat has native footnote numbering for asides without a label (shiny-aside-pill--count, aria-label: "Aside N", index assigned by Dj()), and unlabeled asides are not merged — each gets its own group. That gets real per-citation numbering for free. The trade-off is the accessible name becomes "Aside 1" instead of "documentation"; the source name would live in the popover only.
```

simonpcouch: > One reason I like the pills is that we can give a nice visual clue as to what type of information the citation is referring to through both an icon and a label.

That makes sense! For commons' use case, the type of thing being cited is always a trusted source, so I'm fine with that icon/label being pushed into the hover popup.

> That's fine if you prefer superscripts, but this doesn't feel like the cleanest way to move in that direction. For one, we wouldn't need the image icons that this CSS is referring to now.

I believe they're still needed to be shown on hover, right?

> I think this change also introduces some correctness issues, due to how de-duplicating works in <shiny-aside> that have the same labels.

Thanks, will file an issue!

PR #109, opened 2026-08-13
ci: add package check and address native warning

cpsievert:
## Why this matters

This repository did not previously run `R CMD check` in GitHub Actions. The first commit adds that general package gate, and its initial Linux run exposes an existing compiled-code warning from the sandbox capability probe. The following commits remove that warning and stabilize snapshots against a new DuckDB startup notice.

## What changes

- Add a Linux `R CMD check` workflow for the package suite.
- Preserve the probe child's process-wide termination semantics with `SYS_exit_group`, avoiding the libc `_exit` symbol that the compiled-code scan flags.
- Choose DuckDB's temporary test storage home explicitly, preventing its version 1.5.5 startup announcement from changing error snapshots.

## Stacked follow-up

#107 is stacked on this PR. It introduces the browser-test helper, adds `COMMONS_SKIP_BROWSER_TESTS=true` to the inherited package-check workflow, and runs those browser tests in its own dedicated workflow. This PR deliberately contains no citation-specific browser-test policy.

## Verification

- The workflow-only commit reproduced the `_exit` warning on Ubuntu.
- The subsequent run reported `checking compiled code ... OK`.
- `Rscript -e 'devtools::test(filter = "commons")'` passed all affected snapshot assertions locally; its three remaining failures are the known local macOS FTS extension-download block.\n- The previous fresh Linux run passed the package check, deploy, and pkgdown workflows; this branch rewrite has triggered the final standalone rerun.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -1,0 +1,26 @@
+name: R CMD check
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+permissions: read-all
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v6
+      - uses: r-lib/actions/setup-pandoc@v2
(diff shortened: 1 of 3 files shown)


simonpcouch (approved):
Seems reasonable, thanks!

Issue #91, opened 2026-08-06
make every conversation's calculations reproducible to the app author


simonpcouch:
Halfway there would be inlining the database connection code and then interleaving `run_sql()` calls as `dbExecute()` and `run_r()` as normal R code in a `.R` file.

The proper treatment might use [ir](https://opensource.posit.co/blog/2026-07-23_ir-0-1-0/) based on the active connect deployment and intelligently resolve viewer-based credentials. That still doesn't account for changes in the queried database, though.

PR #111, opened 2026-08-14
fix: make Connect dependencies reproducible

tdstein:
## Summary

- Add a committed renv lockfile and pin the compatible Ellmer/Shinychat revisions.
- Restore the locked environment in CI and generate the Connect manifest from the exact commit being deployed.
- Remove the stale checked-in manifest and limit the generated bundle to production app files.


tdstein:
@simonpcouch, I don't know the R ecosystem very well, so I'm not confident that this is the correct approach. The problem I'm trying to solve is that the original manifest.json kept getting out of date. This change works, but also slows down CI quite a bit.

simonpcouch:
Related to #91. Thank you!

I don't know that renv is necessarily the approach that we want to take here. Related functionality is on the docket for this week under @skaltman's name—I'll defer to her on the right direction to take here.

Issue #56, opened 2026-07-28
Plot / other rich output types for measures


simonpcouch:
Right now, you _can_ output e.g. a plot from a measure but it doesn't get shown in the UI.

simonpcouch:
Closed via #113.

PR #113, opened 2026-08-16
Enable measures to return rich outputs (plots, HTML tables)

skaltman:
Addresses #56. 

I mostly wanted to see what the TLG agent felt like with plots and gt tables supported as measure outputs, so I started working on this. 

Totally fine if we want to go in a different direction for this. 

Idea for more general solution from sol:

>  Future work: Consider generalizing rich_table() into a rich_output() abstraction that
>   separates user-facing HTML, model-facing semantic content, and a calculation-ready
>   run_r value. rich_table() could remain a table-specific wrapper, while the same
>   foundation could support maps, HTML widgets, and other interactive visualizations
>   without applying table-specific rendering or model instructions.

image

image


simonpcouch:
Thank you, @skaltman! I've made some changes here—the big item is removing the exported `rich_table()` helper in favor of automatically showing ggplots and gt tables when they're the tool result of a measure.

Could you install the new version of this PR and then take it for a spin in the TLG agent and let me know if it works for you? You might need to remove any calls to `rich_table()`.

simonpcouch (approved):

skaltman:
@simonpcouch thank you! The plot support works great. The `tlg-agent` complex tables aren't actually `gt_tbl` objects, though, so those measures currently don't work as intended and instead fall back to the generic printed output, which is closed by default. The non-rectangular TLG Catalog tables are `rtables::TableTree` objects rendered as HTML with `rtables::as_html()`. (I realized that my PR description's mention of `gt` was a bit of a red herring.)

Since this is just an example app, I don't think it's that big of a deal to convert them to `gt`, but do you see a different way forward that provides more general support for HTML tables?

I realize that `gt` tables are nice because they carry both the HTML representation and the underlying data frames, so commons can run additional operations on them. Generic HTML support would need to do something different, which is why `rich_table()` coupled the HTML with a separate value.

simonpcouch:
I guess the approach that would allow us to reuse as much of what ellmer/shinychat give us as possible, and also an entry point to model-facing vs. user-facing display, is that measures could return `ellmer::ContentToolResult`s:

```r
  table <- build_table(...)

  ellmer::ContentToolResult(
    value = rtables::as_result_df(table),
    extra = list(
      display = shinychat::tool_result_display(
        html = rtables::as_html(table),
        open = TRUE,
        show_request = FALSE
      )
        data = table
    )
  )
```

Then:

 - the model receives the result data frame;
 - Commons will register `table` for `run_r`
 - shinychat displays the authored rtables HTML.

simonpcouch:
Ah, looks like tlg-agent depended on the old `definitions` syntax and my merge upstream broke it. Working on a fix for tlg-agent locally!

simonpcouch:
Just gave this a go in tlg-agent and it feels good to me! If you could, please take a look + take that for a spin and, if it works well for you, feel free to squash+merge here!

skaltman:
It works great in tlg-agent! Just waiting for the checks to complete...

PR #127, opened 2026-08-18
Restore Shinychat ownership of verified citation display

### LINKED ISSUE #123: Inline citation attachment can break block Markdown and hide verified citations
## Problem

PR #107 originally preserved the whitespace before each verified citation aside.

Commit `f0a46de` added logic that removes this whitespace. The scanner now places
the aside directly after the preceding visible text.

This behavior can break Markdown blocks. A verified citation can disappear from
the user interface, while Commons still records the answer as `Cited`.

## Regression

The last PR commit before the later changes was `5b497232`. For an answer that
ends with a fenced code block, that version produced this structure:

````markdown
```r
1 + 1
```

<shiny-aside ...>
````

After `f0a46de`, the scanner produces this structure:

````markdown
```r
1 + 1
```<shiny-aside ...>
````

The `<shiny-aside>` text becomes part of the code block because the closing
fence is no longer valid Markdown.

The merged state of PR #107, at `426236d`, contains this behavior.

## User impact

I reproduced these results in a real shinychat browser:

- A plain paragraph showed one citation pill.
- A fenced code block showed no citation pill. The aside markup appeared inside
  the code block.
- A Markdown table showed no citation pill. The citation quote appeared as a
  separate blockquote.

The server still records the citation decision as accepted. `stream_async()`
therefore derives provenance tag `B`.

Tag `B` does not get a separate provenance aside. Its citation is expected to
provide the visible trust indicator. The user can therefore see no citation
indicator, while the trace records the answer as `Cited`.

This breaks the contract in PR #107. The displayed provenance and the recorded
provenance no longer represent the same result.

## Root cause

`R/citation-scan.R` now keeps trailing whitespace in
`pending_whitespace`. `emit_attachment()` removes that whitespace when prior
visible text exists:

```r
if (has_attachment_target) {
  pending_whitespace <<- ""
}
```

The scanner does not determine whether the preceding text ends with block
Markdown. It always joins the aside to that text.

## Test gap

The scanner test added with `f0a46de` uses a plain sentence. It asserts that the
aside immediately follows the sentence.

The browser fixture also uses paragraph content. It does not cover answers that
end with:

- A fenced code block
- A Markdown table
- Other block Markdown that requires a line boundary

## Expected behavior

Citation projection must preserve the structure of preceding Markdown blocks.
A verified citation must remain visible when it causes provenance tag `B`.

Add browser regression tests for fenced code blocks and tables. These tests
must also make sure that an accepted citation produces visible citation UI.

cpsievert:
This PR restores the boundary originally established in #107: Commons is responsible for server-side citation verification and audit records; Shinychat is responsible for client-side citation display and interaction.

[Shinychat#326](https://github.com/posit-dev/shinychat/pull/326) adds compact asides (i.e., superscripts), letting Commons request and style compact citations through supported display and CSS APIs rather than reconstructing citation UI with custom markup and private selectors.

## Behavior

- Verified citations render as compact numbered references, with the source label and icon available in the popover and accessible name.
- Citation streaming preserves surrounding Markdown whitespace.
- Commons records citation decisions in the answer audit trail, and trajectory review replays those recorded decisions.

## Verification

Citation scanner and trajectory-review tests cover projection and replay. Browser tests cover streamed compact citations, source-aware markers, and popovers.

Closes #123.


inst/prompts/citation-request.md
```diff
@@ -22,6 +22,4 @@ Rules:
   the quote supports your answer. Put it outside the blockquote.
 - Cite only text that genuinely supports your approach. If nothing you have
   seen supports it, provide no citations.
-- Citations are rendered as footnotes at the end of the preceding line. Place
-  each citation on its own line immediately after the text it supports; that
-  text should stand on its own.
+- Place each citation immediately after the paragraph it supports.
```
simonpcouch: From my understanding of the changes, it doesn't seem necessary to change this line? If so, my more concrete feedback is that I'd prefer not to say "paragraph" here, which might make the agent think it should refrain from writing short replies in favor of writing paragraphs.

R/citation-scan.R
```diff
@@ -20,6 +20,13 @@ ASIDE_OPEN <- "<shiny-aside"
 ASIDE_CLOSE <- "</shiny-aside>"
 ELEMENT_BODY_CAP <- 16384L
 
+# Chunk-invariant incremental scanner: translate model-authored
+# <commons-citation> elements into server-authored <shiny-aside> markup (or
+# drop them, unverified) while copying everything else through unchanged.
+# State spans the unflushed raw-input tail, the text/citation/discard mode, and
+# whether the next raw character is at line start. Text mode retains only a
+# trailing prefix of a reserved literal; citation mode buffers one bounded
+# element body; discard mode drops invalid reserved markup through its close.
```
simonpcouch: Could you do a pass on these comments with a GPT or otherwise non-Claude model, asking the agent to remove all new code comments and then restore only those that speak to the "why" rather than the "what" with clear, concise language? I find Claude code comments to be quite hard to read and often much longer than they need to be.

inst/www/commons-chat/commons-chat.css
```diff
@@ -134,72 +134,24 @@ shiny-chat-container {
...
-  margin-bottom: 0.4em;
-}
-
-.shiny-aside-popover .commons-source-heading img {
-  border-radius: 50%;
-  height: 1.1em;
-  opacity: 0.75;
-  width: 1.1em;
-}
-
```
simonpcouch, line 202: Nice!

inst/www/commons-chat/commons-chat.css
```diff
@@ -134,72 +134,24 @@ shiny-chat-container {
...
+  --shiny-chat-aside-marker-hover-bg: transparent;
+  --shiny-chat-aside-marker-font-family: var(--bs-font-monospace);
+}
+
+/*
+ * Temporary private Shinychat dependency: Shinychat has no public per-aside
+ * tone API, so Commons identifies provenance asides by its own icon URLs.
+ * Remove these selectors when <shiny-aside> gains a public tone contract.
+ * Browser integration tests cover the trusted and warning presentations.
+ */
```
simonpcouch: Helpful to have the public/public boundary documented here, nice.

R/citations.R
```diff
@@ -91,27 +93,24 @@ render_citation_aside <- function(quote, explanation, corpus) {
...
-  sprintf(
-    '<shiny-aside>%s%s%s</shiny-aside>',
-    source,
-    reason,
-    blockquote
+  caption <- paste0(
+    '<small class="text-body-secondary">',
+    "Quote matched exactly against trusted context. ",
+    "Explanation generated by AI.",
+    "</small>"
```
simonpcouch: Let's remove this

R/trajectory-review.R
```diff
@@ -190,6 +190,9 @@ summarize_questions <- function(trajectories) {
   records
 }
 
+# trajectory_read() attaches one recorded provenance entry per reconstructed
+# exchange. Review rendering consumes that audit record rather than re-deriving
+# trust from message text.
```
simonpcouch: Would prefer that these comment-additions-only changes to otherwise unchanged files would be reverted.

R/citation-scan.R
simonpcouch: Relatedly, if you could write a couple sentences in the PR description yourself that help me orient to the changes and understand what you're going for, that would be appreciated! These Claude-generated PR descriptions are hard to parse for me

simonpcouch (changes requested):
Nice! Happy to take this on after some iteration.

Two UI pieces that related PRs and commits have had some back and forth on--in this:

> Image

1) I would prefer that the text in the pills is not bolded. I think the shinychat default weight is 600 and I had changed it to 400.
2) The "Verified answer" badge and text doesn't need to be duplicated inside of the popup


Also, with this PR, I'm seeing that citations often land on separate lines:

> Image

Otherwise, the UI changes are fine with me!

inst/prompts/citation-request.md
cpsievert: Yea, thanks for catching this. Fixed in [9d82a3b](https://github.com/posit-dev/commons/commit/9d82a3b).

R/citations.R
cpsievert: Agreed, fixed in [2adbedd](https://github.com/posit-dev/commons/commit/2adbedd).

R/citation-scan.R
cpsievert: Removed the scanner implementation narration and audited the other new comments. The remaining comments are concise rationale for cross-layer trust, trace-lineage, and temporary Shinychat contracts. Updated in [8bff158](https://github.com/posit-dev/commons/commit/8bff158).

R/trajectory-review.R
cpsievert: Removed the comment-only additions from `trajectory-review.R` in [8bff158](https://github.com/posit-dev/commons/commit/8bff158).

R/citation-scan.R
cpsievert: Rewrote the PR description to summarize the restored Commons/Shinychat ownership boundary, the resulting citation behavior, and the focused verification coverage: [PR #127](https://github.com/posit-dev/commons/pull/127).

cpsievert:
> I would prefer that the text in the pills is not bolded.

Done in [cb33814](https://github.com/posit-dev/commons/pull/127/commits/cb338142eab877bc4c9b354d666a0ebfc6666004)

> The "Verified answer" badge and text doesn't need to be duplicated inside of the popup

Ahh, good catch. For multi-page `<shiny-aside>`, showing the label/icon is usually what you want since they can change, but for single-page it makes sense to drop them. Fixed via https://github.com/posit-dev/shinychat/pull/326/changes/16df7d51cf5559225c03e6390c8cf878ae090c72

> I'm seeing that citations often land on separate lines

Another good catch. `<shiny-aside>` already has some heuristics to bring the asides inline to the preceding paragraph / list item when it makes sense to. It wasn't capturing a case like this though: `Text\n\n<shiny-aside>`. Addressed via https://github.com/posit-dev/shinychat/pull/327, but please lmk if you encounter more situations like this!

inst/prompts/citation-request.md
```diff
@@ -22,6 +22,4 @@ Rules:
   the quote supports your answer. Put it outside the blockquote.
 - Cite only text that genuinely supports your approach. If nothing you have
   seen supports it, provide no citations.
-- Citations are rendered as footnotes at the end of the preceding line. Place
-  each citation on its own line immediately after the text it supports; that
-  text should stand on its own.
+- Place each citation immediately after the text it supports.
```
simonpcouch:
```suggestion
- Citations are rendered as footnotes at the end of the preceding line. Place
  each citation on its own line immediately after the text it supports; that
  text should stand on its own.
```

Without this further bit of context, I'm seeing that the agent sometimes ends a line of text with a colon or an incomplete fragment. Let's keep this piece of prompting as it was.

simonpcouch (approved):
Looks great, thanks!

[edit: there had previously been a screenshot here where i said there was still a newline before citations. i was wrong--didn't have the most recent shinychat. that's fixed!]

Otherwise, looks good. I trust you to make changes as needed and merge when the changes are ready!

inst/prompts/citation-request.md
cpsievert: Oh geez, sorry, [9d82a3b](https://github.com/posit-dev/commons/commit/9d82a3b) was not the right fix. Thanks for catching (again)

PR #143, opened 2026-08-21
degrade gracefully without a run_r sandbox

### LINKED ISSUE #142: consider disabling `run_r` rather than `abort()`ing deployment when sandbox is unavailable
Some small percentage of systems won't be able to use either Landlock or the fallback sandbox. Right now, we just refuse to deploy in that case. One alternative is to just deploy without the `run_r` tool and raise a warning to the app author linking them to a help page that explains their situation. (The folks in that situation have an old Linux _and_ have changed the default `user.max_user_namespaces` to 0.)

simonpcouch:
Closes #142. Instead of failing, we just return a harness without `run_r` (and provide a warning).

Was initially unsure that I wanted to go down this road, but the implementation is less complex than I thought it'd be. I haven't yet tested this interactively.


simonpcouch:
Let's not do this for now.

PR #144, opened 2026-08-21
some design tweaks

### LINKED ISSUE #141: make `commons_ui()` feel more like a modern chat app
The current UI still somewhat "feels like shiny / bslib." shinychat has to be very general and also supports a broad existing user base; commons is a batteries-included version that should just feel like a modern, 'normal' chat app to its end users.

What design elements could we modify from shinychat that would make the interface feel more modern and sleek? For one, shapes and rounding should be more consistent. Another thought is that the conversation selector in other modern chat apps is always-on (or at least on-by-default) and doesn't grey out the current conversation. Those apps tend to make more use of glimmer/shimmer to indicate progress, too. We could also further refine the landing page experience especially, with a larger and more centered input box and animation that sends it to the bottom of the page after (shinychat already does this with `greeting` with `suggestion`s).

simonpcouch:
Closes #141.

Before:

https://github.com/user-attachments/assets/df5a16e1-0d9a-41ac-bb3b-c65df0c995d8

With this PR [EDIT: some further tweaks visible in a screenshot below]:

https://github.com/user-attachments/assets/5c440d9f-801e-41ef-a54d-15c89916d464

### DIFF
--- a/pkg-r/inst/www/commons-chat/commons-chat.css
+++ b/pkg-r/inst/www/commons-chat/commons-chat.css
@@ -5,39 +5,410 @@
 /* ---- Chat-level polish ---------------------------------------------- */
 
 shiny-chat-container {
+  --commons-chat-radius-control: 0.5rem;
+  --commons-chat-radius-surface: 0.75rem;
+  --commons-chat-radius-message: 1.25rem;
+  --commons-chat-radius-round: 999px;
+  --commons-chat-motion-fast: 120ms;
+  --commons-chat-motion-base: 180ms;
+  --commons-chat-ease-out: cubic-bezier(0.2, 0, 0, 1);
+  --commons-chat-state-hover: color-mix(
+    in srgb,
+    var(--bs-emphasis-color, #000) 6%,
+    transparent
+  );
+  --commons-chat-state-active: color-mix(
+    in srgb,
+    var(--bs-emphasis-color, #000) 10%,
+    transparent
+  );
+  --commons-chat-border-strong: color-mix(
+    in srgb,
+    var(--bs-body-color, #212529) 24%,
+    transparent
+  );
+  --commons-chat-focus-color: var(--bs-primary, #007bc2);
+  --commons-chat-focus-ring: color-mix(
+    in srgb,
+    var(--bs-primary, #007bc2) 16%,
+    transparent
+  );
+  --commons-chat-shadow-control:
+    0 1px 2px color-mix(in srgb, var(--bs-body-color, #212529) 8%, transparent),
+    0 8px 24px -20px
+      color-mix(in srgb, var(--bs-body-color, #212529) 30%, transparent);
   --shiny-chat-user-message-bg: var(--bs-tertiary-bg, #f4f5f7);
-  --shiny-tool-card-spinner-color: var(--bs-secondary-color, #6c757d);
+  --shiny-chat-streaming-color: var(--bs-primary, #007bc2);
+  --shiny-tool-card-spinner-color: var(--bs-primary, #007bc2);
 }
 
-.shiny-chat-user-message {
-  border-radius: 1.25rem;
-  padding: 0.65rem 1.1rem;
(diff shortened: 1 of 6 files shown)


skaltman:
The `Working...` functionality is really nice. I also like that the stop button is now neutral instead of red. 

Generally, the spacing and shape feel like they are moving in two directions. The message composer got taller, while the user messages got more tightly wrapped. I think it would feel more coherent if everything moved in the same direction: more or less padding, more or less boxy. I think I would prefer more horizontal padding and slightly boxier. I like how Posit Assistant's user messages feel, which I think have more horizontal padding than shinychat's.

* The chat composer feels somewhat too tall and clunky, caught between the more compact shinychat pill version and a more expansive, boxier feel. I think it should just move slightly in either direction to feel more intentional. 
* The send message button still feels very shinychat-like and prominent to me. What do you think about making it hollow, like Posit Assistant? 
* Changing the user message background color might also help it feel less default-shinychat. It could stay gray, but be a different shade, possibly a light blue-gray? I don't feel strongly about this.

skaltman:
Also, do you think that the send-button spinner is necessary? Could the send button just be disabled and gray during that period instead? This might fall under shinychat work, though.

simonpcouch:
Here's another take on those shapes!

> Screenshot 2026-08-21 at 1 35 27 PM

The default shinychat color is actually a light blue and we had changed it to a light grey. Here's the default blue with the new geometry:

> Screenshot 2026-08-21 at 1 37 16 PM

I actually think the default shinychat blue looks fine here!

skaltman:
This looks nice! I like the shapes in this version

simonpcouch:
@gadenbuie, do you mind to take a look here and share any reactions? Did my best not to overlap with `page_chat()` changes and broadly not step on internal shinychat toes.

The "Working..." change specifically might be of interest for shinychat, as an alternative to the spinner icon. The implementation on the shinychat side could be cleaner than the one here, where we use a constant delay to avoid a flicker.

pkg-r/inst/www/commons-chat/commons-chat.css
```diff
@@ -5,39 +5,488 @@
...
+shiny-chat-container .shiny-chat-input .shiny-chat-btn-attach:disabled {
+  opacity: 0.38;
+}
+
+/* Match shinychat's activity-row geometry so pending and thinking stay aligned. */
+shiny-chat-container
+  .shiny-chat-message:not(:has(> .message-icon))
+  > .shiny-chat-message-content:empty {
+  --_row-pad: 0.4rem;
+  --_row-glyph-width: 0.9em;
```
simonpcouch: Ah, these are treated as private in shinychat.

pkg-r/inst/www/commons-chat/commons-chat.css
```diff
@@ -5,39 +5,488 @@
...
+  --commons-chat-focus-ring: color-mix(
+    in srgb,
+    var(--bs-body-color, #212529) 10%,
+    transparent
+  );
+  --commons-chat-shadow-control:
+    0 1px 2px color-mix(in srgb, var(--bs-body-color, #212529) 8%, transparent),
+    0 8px 24px -20px
+      color-mix(in srgb, var(--bs-body-color, #212529) 30%, transparent);
+  --shiny-chat-user-message-border-radius: var(
```
simonpcouch, line 38: These use new variables from page_chat.

gadenbuie:
I fixed the missing visual feedback in shinychat when waiting for the API response to start streaming in when you don't have an icon, although I went a slightly different direction that re-uses some existing shinychat design. Also the assistant icon is now disabled by default in dev shinychat

* https://github.com/posit-dev/shinychat/pull/345

simonpcouch:
Thank you @gadenbuie! Had just been poking at this again but will wait and see how posit-dev/shinychat#341 comes together before merging.

simonpcouch:
Closing in favor of the two new linked issues!

Issue #138, opened 2026-08-21
audit package-level documentation


simonpcouch:
* README
* DESCRIPTION
* Getting started vignette (h.t. @skaltman #137)
* Security vignette
* pkgdown function reference
* Options and envvars (there should only be a couple user-facing ones IIRC, rest are dev)

simonpcouch:
Related to #152, #150, d05c6e2.

After merge of Sara's README changes, I think the only remaining item is auditing + documenting options and envvars.

skaltman:
Very minor related issue: https://github.com/posit-dev/commons/issues/157 

There might be a handful of similar standardization issues, it might be worth a pass.

PR #146, opened 2026-08-21
Restructure as an R + Python monorepo

jat255:
Moves the R package from the repository root into `pkg-r/`, adds a `pkg-py/` skeleton, and establishes `tests/shared/` for cross-language spec fixtures. Implements kata `z26z` / M1 Task 0 of the Python port plan (decision D7).

**Includes three kata-config commits alongside the restructure**, adding `.kata.toml`, a kata section in `AGENTS.md`, and one `.gitignore` entry. They are independent of the move and touch three files in total.

## Why one repository

commons' behavior lives substantially in artifacts that have to agree across both languages: the system prompt, the citation dialect and its guards, the provenance truth table and its display copy, and the tracing span contract. In separate repositories each of those is a copy that drifts silently, and prompt drift fails invisibly, since nothing errors and the agent simply behaves differently in one language. `tests/shared/` is where those contracts become fixtures CI enforces instead of prose that rots. The precedent is [posit-dev/shinychat](https://github.com/posit-dev/shinychat).

## Reviewing this

The first commit is a pure relocation, verifiable rather than taken on trust: `git show --stat -M` reports `172 files changed, 0 insertions(+), 0 deletions(-)`. The remaining commits are each a single concern (CI scoping, install instructions, root scaffolding, docs), so reviewing commit by commit is much easier than reading the squashed diff.

The repository is knowingly broken at the first commit, since the workflows still point at root-relative paths until the third.

## Things a reviewer could not infer from the diff

- **The move silently broke five `.gitignore` patterns.** A pattern containing an internal separator is anchored to the directory of the file declaring it, so `inst/resources/`, `inst/tiles/`, `inst/pkgs`, `inst/hex/output` and `src/*.{o,so,dll}` all stopped matching once the package moved. The failure is quiet: compiled objects and generated asset directories just become committable. Fixed by splitting package-relative patterns into `pkg-r/.gitignore` and leaving only separator-free patterns at the root, where they match at any depth and so cover `pkg-py/` too. `*.Rcheck/` and `*.tar.gz` are also now ignored, which they never were before the split.
- **Two workflows needed changes that the issue does not mention.** `deploy.yml` pushed `path: inst` to Connect and would have kept reporting success while deploying a path that no longer exists; it now uses `pkg-r/inst` and has a `paths:` filter so a Python-only PR cannot trigger an R deploy. `citation-browser.yaml` runs `devtools::test()` and needed the same working-directory treatment as `R CMD check`.
- **`defaults.run.working-directory` does not apply to `uses:` steps**, so the r-lib actions take their own `working-directory` input. `R-CMD-check` has no `run:` steps and therefore no job default at all. pkgdown's deploy step keeps `folder: pkg-r/docs` spelled from the repository root, because the JamesIves action resolves it there rather than from the job's working directory.
- **`NEWS.md` does not exist**, despite being named in the issue's move list.
- **The `subdir` sweep was two occurrences**, not the wider sweep across vignettes and `inst/skills/` the issue anticipated. Spelled as pak's path form `posit-dev/commons/pkg-r` rather than `subdir = "pkg-r"`, which is `remotes::install_github()`'s argument.

## Deliberate departures

- **`pkg-py/` is not empty.** The task asks for py-check green against an empty directory, which is not possible: `uv sync` needs a `pyproject.toml` and `pytest` exits 5 on a suite with no tests. It carries a placeholder `pyproject.toml` (`package = false`) and one placeholder test, both of which the scaffold task deletes. A workflow green because it does nothing would be worse than no workflow.
- **`LICENSE` and `LICENSE.md` are duplicated** at the root and in `pkg-r/`. `R CMD check` resolves the `DESCRIPTION` reference inside the package, and GitHub only detects a repository license at the root.
- **Split into 12 commits** rather than the single commit the task specifies, so each is independently reviewable.

## Needs a repository admin, not a code change

`paths:` filters mean a required status check that never runs stays pending forever under branch protection and blocks the merge. If any R check is currently required, that setting needs adjusting before this lands.

## Verification

py-check verified locally: `uv sync --locked`, `ruff`, `pytest` all pass. `R CMD build` from `pkg-r/` produces a valid tarball and the structural checks the move could have broken (`DESCRIPTION` located, namespace information) pass.

`R CMD check` and the pkgdown build could not be completed locally, since none of the package's twelve hard dependencies are installed in the available R library. That is the one item this PR cannot self-certify, and CI covers it.

This branch was rebased onto three commits that landed upstream after it was written (`#132`, `#138`, `#145`). Those added five new files under `R/` and `tests/testthat/`, which git's directory-rename detection folded into the relocation commit; all 27 files those commits touched are byte-identical to `main` inside `pkg-r/`.

## Not included

`tests/shared/` holds only the contract describing what belongs there. Fixtures land as the provenance and citation code is ported, together with the R-side runner. The Python package itself is the next task.

Shared *shipped* artifacts are also out of scope here and decided separately as plan D10: the system prompt moves to a root `prompts/` source in M5 and the browser assets to a root `www/` in M8, each synced into both packages with CI failing on a stale copy. Worth knowing because neither an R package nor a Python wheel can ship a file from outside its own directory, so sharing always means a generated copy rather than a direct read. That is also why `tests/shared/README.md` describes the R suite as consuming a synced copy.

`MIGRATING.md` covers what a developer used to R tooling at the repository root needs to change. It should be removed at a later date once we're all comfortable with the monorepo.


jat255:
The failing `R CMD check` here is inherited from `main` and not caused by this PR.

`main`'s own run at `2221b88` fails with a byte-identical signature, `FAIL 8 | WARN 0 | SKIP 27 | PASS 1734`, and the same error: [main's run](https://github.com/posit-dev/commons/actions/runs/32498977049) against [this PR's run](https://github.com/posit-dev/commons/actions/runs/32514148357). This branch touches no test code and no `DESCRIPTION`; the eight failing tests are files it only relocated.

The cause is #145. Removing the `Remotes: tidyverse/ellmer` entry means CI now installs released ellmer rather than the development version, and the eight tests in `pkg-r/tests/testthat/test-commons.R` that call `local_mocked_bindings(..., .package = "ellmer")` mock two ellmer internals that only exist on ellmer's development branch. `stream_content_with_turns` and `value_turn_with_turns` are both present in `R/provider.R` on ellmer `main` but absent at tag `v0.4.1`, so the mocks now fail with `Can't find binding`.

This failure mode arrived with #145. `ff184d5` was also red, but for a different reason (`FAIL 4 | SKIP 14 | PASS 1581`).

Nothing in this PR can turn that check green, so it needs fixing on `main`. Three options, and the middle one looks best to me because it matches the guidance in `AGENTS.md` about avoiding excessive mocking:

1. Restore the `Remotes: tidyverse/ellmer` entry, which gives up what #145 set out to do.
2. Rewrite the eight tests so they do not reach into ellmer's internals.
3. Skip them when the bindings are absent, which keeps the coverage gap invisible.

Worth deciding before this merges, since a reviewer otherwise has no way to tell the inherited failure apart from a regression introduced by moving 172 files.

MIGRATING.md
simonpcouch, line 1: Appreciate you putting this together!

README.md
simonpcouch, line 1: I'm hesitant about this README rewrite—I do think it's important for passersby to see human-written text and   the visual demo, and I think we can just make a few tweaks to the existing one to say e.g. "chatlas and ellmer" and "To install, run xyz in R or abc in Python." Fine to copy the existing one into pkg-r and have shared content there, as that content will end up as the index page for the automatically generated site!

Fine with me that the developer-facing content here would really only live in AGENTS.md.

pkg-r/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+inst/tiles/
+inst/tiles-agent/
+inst/pkgs
+inst/hex/output
+src/*.o
+src/*.so
+src/*.dll
+
+# R CMD build / check output. Never ignored before the split either, so this
+# closes a pre-existing gap rather than restoring something the move broke.
```
simonpcouch:
```suggestion
```

pkg-r/.gitignore
```diff
@@ -0,0 +1,21 @@
+# Package-relative ignores.
+#
+# These live here rather than in the repository root because a gitignore
+# pattern containing a separator is anchored to the directory of the
+# .gitignore that declares it. `inst/resources/` and `src/*.o` at the root
+# stopped matching the moment the package moved under pkg-r/, which would
+# have let compiled objects and generated asset directories become
+# committable without anyone noticing.
```
simonpcouch:
```suggestion
```

I think these .gitignore entries are self-explanatory

simonpcouch (approved):
Nice, looks good! I appreciate you taking care of this as thoughtfully as you have.

Only substantive comment is about README.md; I really have a thing about repo README aesthetics😅😭

CODEOWNERS
```diff
@@ -0,0 +1,15 @@
+# Review routing per language. Paths are matched last-rule-wins, so the
+# default owners come first and the more specific paths override them.
+
+*                   @simonpcouch @skaltman
+
+/pkg-r/             @simonpcouch @skaltman
```
simonpcouch:
```suggestion
```

I do wonder whether this is a week premature. At least on the R side, I think we'll need to continue making some changes fast and loose. I trust you to request review on things that need review, and myself will need the flexibility to make small, trivial fixes on the R side.

pkg-r/.gitignore
jat255: Done in `80e6761` — both comment blocks are gone and the file is just the ten patterns now. Agreed the entries speak for themselves; what I was documenting was why the file exists in `pkg-r/` rather than at the root, and that's preserved in `0948cc6`'s commit message and in MIGRATING.md. Re-checked that all ten patterns still match after the edit, since a stray deletion there would be silent.

pkg-r/.gitignore
jat255: Done in `80e6761`, same edit as the block above.

CODEOWNERS
jat255: Removed the whole file in `80e6761` rather than just this line. We can put these back in place once things have stabilized.

README.md
jat255: Worth knowing that this thread is anchored to `fc5b099`, and I rewrote the README about four hours after you left it, so what you were reading is not what is there now. I think the current version is close to what you are describing, and it would be good to know whether you still object to it.

It keeps the human-written prose, the hex logo, and the screencast, reusing the text from the existing README rather than replacing it. `pkg-r/README.md` is untouched by the move apart from the install line, so it still works as the pkgdown index page. Developer-facing content is out of it entirely as of `80e6761`: the package-directory rule and the MIGRATING.md pointer now live only in AGENTS.md, which I think is the split you were after.

Two of your asks I handled differently, and I would rather explain than quietly not do them.

chatlas is now named next to ellmer, but phrased as "once the Python package is ready" rather than presented as current. Writing it flat as "chatlas and ellmer" reads as though Python support exists, and `pkg-py/` is a placeholder `pyproject.toml` and one placeholder test.

For "to install, run xyz in R or abc in Python", there is no Python command to give: nothing is published to PyPI yet. The R section has the real `pak::pak()` call and the Python section says the package is in development and not yet installable, with the `posit-commons` / `import commons` naming so the eventual shape is visible.

Happy to go further in either direction if you would still rather see the two languages presented symmetrically.

jat255:
Thanks @simonpcouch. All four points addressed in 80e6761, with details on the individual threads.

A couple notes that differed from your suggestions: **CODEOWNERS** is deleted entirely rather than losing just the `/pkg-r/` line, to facility rapid changes (we can add it back later on). I edited the top-level **README** to keep the prose, logo, and screencast and be user-facing rather than developer-facing.

The `.gitignore` comments are gone, and the developer-facing content now lives only in `AGENTS.md`.

I also merged in latest main.

README.md
```diff
@@ -1,45 +1,36 @@
...
-existing work, like Shiny apps and Quarto docs, to answer questions.
-(You can also import existing trusted calculations from semantic layers
-in Snowflake and Databricks.) When the question can’t be answered by a
-trusted calculation, the agent can search across context you’ve compiled
-to query data directly, and the response will be deterministically
-tagged as untrusted.
+## Two languages, one design
 
-<img src="https://github.com/user-attachments/assets/67dcf1f2-1496-406a-96cb-50a2e7050eeb" alt="A screencast demonstrating a commons data agent answering questions with a trusted calculation and then a direct data query. In the first case, there's a provenance pill that marks the answer as verified. In the second case, the pill reads 'Untrusted.'" width="100%" />
+This repository holds both implementations of commons. Today only the R package works. The Python package is being built to the same design, with the same three layers (data, semantic, and context), the same tools, and the same provenance semantics, so that an agent will eventually behave the same way whichever language you build it in.
```
simonpcouch: I think we can essentially write the README aspirationally? Since there's an experimental badge, I think we're covered in just writing that things work and getting there eventually.

Fine with me if you decide to ignore here!

simonpcouch (approved):

README.md
jat255: Done in [`8723f18`](https://github.com/posit-dev/commons/pull/146/commits/8723f18c4f9c0c4536ceab71c5f10de3ea7ec94c). This was caused by roborev being unnecessarily pedantic about the README not matching the current state of the code (which I suppose is good in general, but annoying here...)

README.md
simonpcouch, line 1: Noting that the merge upstream now revert Sara's changes from #152.

jat255: They got moved to `pkg-r/README.md`:

image

README.md
```diff
@@ -1,107 +1,43 @@
...
 
 <img src="https://github.com/user-attachments/assets/67dcf1f2-1496-406a-96cb-50a2e7050eeb" alt="A screencast demonstrating a commons data agent answering questions with a trusted calculation and then a direct data query. In the first case, there's a provenance pill that marks the answer as verified. In the second case, the pill reads 'Untrusted.'" width="100%" />
 
-## Installation
+commons agents support a wide variety of LLM providers, via [ellmer](https://ellmer.tidyverse.org/) in R and [chatlas](https://posit-dev.github.io/chatlas/) in Python. Extracted context is stored in plain-text [data-dict.yaml](https://data-dict.tidyverse.org/) alongside `.R` or `.py` files.
+
+## Two languages, one design
+
+This repository holds both implementations of commons. They share one design: the same three layers (data, semantic, and context), the same tools, and the same provenance semantics, so an agent behaves the same way whichever language you build it in. Pick the one your team already works in.
```
simonpcouch, line 20: I don't know that this paragraph would make sense to someone who doesn't already know what the package does

README.md
simonpcouch: Yes, though I think many of her changes are relevant for this main README!

simonpcouch (approved):

PR #151, opened 2026-08-21
Fix Getting Started vignette verification pill rendering

skaltman:
The pills are currently not rendering correctly. This fixes the issue by registering the existing commons chat dependency with the vignette.


simonpcouch (approved):
🏄

PR #152, opened 2026-08-21
Add details to README

skaltman:
Updates and adds more detail to the `README`, including a section on the eval results from DevRel Agent. 

* I can update the gif when the commons UI settles. 
* The DevRel agent data is temporarily hardcoded. I opened [an issue](https://github.com/posit-dev/devrel-agent/issues/6) to include a file with the results so we can read in the data here. 

image


man/figures/README-eval-plot-1.png
simonpcouch, line 1: Really sharp, I dig this plot.

README.md
```diff
@@ -42,4 +33,73 @@ To install the package, run:
...
+the agent.
+
+Answers are tagged according to the analysis path followed, so users can
+determine how much trust to put in a given answer.
+
+For more information, see the [Introduction to
+commons](https://posit-dev.github.io/commons/articles/commons.html)
+vignette.
+
+<img src="man/figures/README-trust-flow.png" alt="A question first searches trusted calculations. The high-trust path runs a relevant trusted calculation and produces a verified answer. The lower-trust path searches context and writes custom SQL or R, producing either a cited or untrusted answer." width="100%" />
```
simonpcouch, line 85: Is there source code for this figure somewhere? (The Getting Started vignette, I'm guessing?) Could we include a link to it in an HTML comment?

simonpcouch (approved):
Just one small reproducibility ask, otherwise looks great.

Linking to #146, which also makes some README edits. Whichever one merges second will need to reconcile edits from the other.

README.md
skaltman: There is [already a comment](https://github.com/posit-dev/commons/pull/152/changes#diff-72778b58969c8ca8268402860b0e003e3d213a26c812bc9f9b928395c284c99fR85)! 

I can put it in HTML though so it shows up in the `.md` as well.

simonpcouch: ooooomg, looking at the wrong file. My bad!

skaltman:
@simonpcouch I just did some minor edits, so I will merge now!

PR #163, opened 2026-08-25
feat: Implement provenance in Python behind a shared fixture

jat255:
## Summary

Adds `derive_provenance_tag()` and the provenance display copy to the Python package as `commons._provenance`, with `Tag` as an enum rather than bare strings. 

The truth table and the display strings are behavior both packages must agree on, so they live in `tests/shared/provenance.json` and both suites enumerate the same cases rather than each restating them. The R suite reads a generated copy, because testthat needs its fixtures inside the package and an installed package cannot reach the repository root. `scripts/sync-shared-fixtures.sh` produces that copy and a new `verify-shared-fixtures` workflow re-runs the script and fails when the committed copy is stale.

## Notes for the reviewer

**This PR touches R code.** This PR lands the first shared fixture and the mechanism around it. That mechanism is the deliverable to weigh most on the R side.

Python notes:

**`Tag` is a `StrEnum`, not a plain `str, Enum` mixin.** That mixin formats as `"Tag.A"`, while R writes the bare letter to the `commons.provenance.tag` span attribute. An f-string would have produced Python traces the R trajectory reviewer cannot match, and nothing would have errored. There is a test pinning `str(Tag.A) == "A"`.

**`pkg-py/pyproject.toml` gains `search-path` for pyrefly.** `tests` is a package rooted at `pkg-py/`, not under `src/`, so the import between test modules did not resolve and `py-check` would have failed. Worth knowing for future work: running `pyrefly` from inside a `.worktrees/` checkout reports a false pass, because it silently omits `tests/` whenever any ancestor directory name begins with a dot. Verified by reproducing the same tree at a path with and without a dot-directory. Local type checking of test files has to happen somewhere outside `.worktrees/`.

**`AGENTS.md` gains a writing rule, and this branch applies it.** Neither package is a port of the other, so the Python package's docstring, its PyPI `description`, and the kata section no longer say it is. Only the docstring came from this branch; the other two predate it, and landing the rule while leaving them would have contradicted it on arrival. Instances of "port" that describe R code carried over from data-dict are left alone, since those are accurate and describe the opposite direction.

**`provenance_aside()` has no Python counterpart.** It has no caller there yet, and depends on icon resolution that does not exist there either.

The fixture only covers tags a tool can actually set. `pkg-r/R/tools.R` emits A and B; C is derived and never an input, so no case feeds it in. Both runners assert the case list is non-empty, because a truncated fixture would otherwise enumerate nothing and pass.

## Testing

ruff, pyrefly, and pytest pass on Python 3.11 and 3.13. The two new R tests pass against the real `R/provenance.R`.

I checked that the fixture actually governs both languages rather than just sitting beside them: renaming one display label in the source makes the R test fail, the Python test fail, and the sync check go dirty.

Full `R CMD check` was not run locally, since commons' Imports are not installed here. CI covers it.

One non-blocking observation, not addressed here: `MIGRATING.md:53` enumerates the CI workflows as a complete list, and has been out of date since #161 added `py-release`.


simonpcouch (approved):
This seems reasonable to me!

PR #165, opened 2026-08-25
feat: Add citation normalization and matching to the Python package

jat255:
## Summary

This PR adds citation normalization and matching to the Python implementation, via the `commons._citations` module (the equivalent of `R/citations.R`). The normalization rules and the matching verdicts are shared, live in `tests/shared/citations.json`, and both the Python and R suites run the same contract.

The Python implementation uses a `CorpusEntry` dataclass to hold an immutable record of one trusted passage a quote can be checked against, holding the passage itself (`text`), a reader-facing `label` that names its source in the rendered aside, and a `kind` of `"prose"`, `"definition"` or `"schema"` saying which layer it came from.

## Changes affecting R

`pkg-r/tests/testthat/test-citations.R` loses two hand-written tests, `"citation matching forgives reflowed whitespace and typography"` and `"trivial quotes cannot promote an answer"`. Both stated expectations the Python suite also needs, so they move into the shared fixture using the machinery from #163. Three fixture-driven tests replace them, covering the normalization rules, the matching verdicts, and the minimum quote length.

## Shared-fixture guidance

`AGENTS.md` and `tests/shared/README.md` are updated in this PR to clarify that the shared fixtures should govern behavior a user can observe rather than specific implementation detail. The goal here is that each package should implement idiomatically in its own language.

## Testing

The fixture holds 12 normalization cases and 11 matching cases, including both guards: the 10-character minimum, tested at its boundary and with a case proving it counts the normalized form rather than the raw input, and the rule that only the quote is ever verified, which is structural here since `match_citation()` takes a quote and a corpus and nothing else. Both suites enumerate those cases.


pkg-r/R/citations.R
```diff
@@ -52,10 +52,13 @@ build_citation_corpus <- function(context_layer, registry, sources) {
   corpus
 }
 
+# A quote this short can occur in a trusted source by coincidence, so it must
+# not promote an unsupported answer. Enforced/tested by tests/shared/citations.json.
+CITATION_MIN_NORMALIZED_LENGTH <- 10
```
simonpcouch: There are admittedly some exceptions to this in commons, but tidyverse style is [lower snake case, even for constants](https://style.tidyverse.org/syntax.html#object-names)

simonpcouch: I also find this reworded comment quite hard to read. Could you just bring over the existing one?

pkg-r/R/citations.R
```diff
@@ -105,14 +108,28 @@ citation_aside_html <- function(quote, explanation, label, kind) {
...
 # Forgiving of the ways a faithful quote can still drift from its source:
 # reflowed whitespace, markdown emphasis, and typographic quotes/dashes.
 normalize_citation <- function(x) {
   x <- gsub("[*_`]", "", x)
   x <- gsub("[\u2018\u2019]", "'", x)
   x <- gsub("[\u201c\u201d]", "\"", x)
   x <- gsub("[\u2013\u2014]", "-", x)
-  trimws(gsub("\\s+", " ", x))
+  # Collapsing first leaves at most a single space at each end, so trimws()
+  # never has to consider anything outside its default set.
```
simonpcouch:
```suggestion
```

pkg-r/R/citations.R
```diff
@@ -105,14 +108,28 @@ citation_aside_html <- function(quote, explanation, label, kind) {
   )
 }
 
+# To share an implementation contract with the Python side, we need to
+# explicitly reference the Unicode White_Space set. Due to inconsistencies
+# between R and Python, we intentionally don't use either engine's `\s`.
+# On R, the default TRE engine defers to locale-dependent C library classification,
+# which disagrees with Python about a non-breaking space and may disagree with itself
+# across platforms. This behavior is enforced/checked by tests/shared/citations.json.
```
simonpcouch:
```suggestion
# Spell out Unicode whitespace for consistency across R, Python, and locales.
```

pkg-r/tests/testthat/test-citations.R
```diff
@@ -1,18 +1,76 @@
...
+    )
+  }
+})
+
+test_that("match_citation uses the shared minimum quote length", {
+  # The fixture declares the minimum as a value, not only through the boundary
+  # cases, so both implementations must read the same number from it.
+  minimum <- shared_fixture("citations")$match_citation$min_normalized_length
+  # Size the corpus from the minimum, so raising it cannot make the acceptance
+  # fail for absence rather than for disagreeing with the fixture.
```
simonpcouch: I'm really struggling to wrap my head around what this means

simonpcouch (commented):
Added some comments, but from a broader perspective, I'm wondering whether this is actually necessary. This feels like quite a bit of machinery for a relatively trivia-flavored edge case? Is this likely to bite us in practice?

This specific shared fixture would be fine on its own, but I'm hesitant to set the expectation that the two implementations remain _this_ tightly aligned on implementation details.

jat255:
> Added some comments, but from a broader perspective, I'm wondering whether this is actually necessary. This feels like quite a bit of machinery for a relatively trivia-flavored edge case? Is this likely to bite us in practice?

Taking a second glance at it, you're probably correct here (though maybe non-english sources could have more Unicode weirdness -- that's a stretch though). 

> This specific shared fixture would be fine on its own, but I'm hesitant to set the expectation that the two implementations remain _this_ tightly aligned on implementation details.

I hear you on this. If some divergence in behavior is fine, I think it's likely better to stick with what is idiomatic for each language (such as, in this case using `\s` for each, even if that means slightly different results, potentially). I'll rework this to just use that approach instead.

simonpcouch (approved):

pkg-r/tests/testthat/test-citations.R
jat255: Hah, that one slipped through when I was reviewing the comments. Good work on that one Claude...

PR #166, opened 2026-08-25
transition from pills to badges

### LINKED ISSUE #148: consider switching to the default shinychat citation UI
Carson has introduced some CSS variables where we could style the background/border/text color/weight.

### LINKED ISSUE #149: consider switching to little superscript dots for Verified/Untrusted
Instead of Verified/Untrusted pills, we could add little superscript dots (or maybe styled like familiar signs, like Stop / Caution for the yellow).

> Image

In this case, we would stick with the minimal citations look that we currently use. (re: #148)

simonpcouch:
Closes #149 (and closes #148 as wontfix). Transitions from pills (and footnote syntax for cited responses) to a small badge. Green > Blue > Yellow, as a little icon rather than dot for accessibility. For the first second or so, the badge pulses.

> Screenshot 2026-08-25 at 12 29 33 PM


simonpcouch:
Closing in favor of #180.

PR #169, opened 2026-08-25
Adopt shinychat's page_chat(): replace commons_ui() with commons_theme()

cpsievert:
shinychat's new [`page_chat()`](https://github.com/posit-dev/shinychat/pull/329) owns page composition, which left `commons_ui()` — our thin `chat_ui()` wrapper that attached commons's CSS/JS — with no way to get those assets onto the page. This PR removes `commons_ui()` and moves asset delivery into `commons_theme()`, an ordinary bslib theme that works with `page_chat()` or any bslib page. `commons_server()` is unchanged.

The encouraged pattern for a commons app is now plain shinychat UI plus the theme:

```r
library(shiny)
library(shinychat)

ui <- page_chat("Assistant", id = "chat", theme = commons_theme())

server <- function(input, output, session) {
  agent <- commons(
    ellmer::chat_anthropic(),
    data_sources = data_source(sales = sales)
  )
  commons_server("chat", agent)
}

shinyApp(ui, server)
```

For an embedded chat, use `chat_ui()` inside any bslib page with the same `theme = commons_theme()`; `inst/app.R` is a complete example.

One thing worth knowing when reviewing: theme dependencies render *before* shinychat's own stylesheet, so commons's `.shiny-chat-*` overrides needed strictly higher specificity (or, where shinychat exposes a theme variable, moving to the variable instead). Caught via live browser testing; details are in the CSS and `commons_theme()`.

Commit walkthrough:

- e85af0b — add `commons_theme()`.
- 978b4f1 — remove `commons_ui()`; port `commons_app()`, the trajectory reviewer, examples, tests, and docs.
- a1e566f — drop `icon_assistant = HTML("")`; shinychat now omits the icon by default ([#345](https://github.com/posit-dev/shinychat/pull/345)).
- e934e97 — drop explicit `enable_cancel`/`allow_attachments`; `NULL` defers to `chat_server()` ([#264](https://github.com/posit-dev/shinychat/pull/264)).
- 45bee16 — serve the icon SVGs from the `commons-chat` dependency instead of a registered resource path.
- d20d254 — move bslib/htmltools/sass to Imports (already transitive via shinychat).
- 9123484 — drop a redundant example app; `inst/app.R` covers the pattern.

Verified: the four affected test files pass, and the example apps were exercised live in a browser with a real agent (assets load via the theme, pills/tool rows render with commons styling, icons serve 200). The full suite and shinytest2 tests were too slow locally — relying on CI for those.


pkg-r/R/theme.R
```diff
@@ -0,0 +1,51 @@
...
+#'     ellmer::chat_anthropic(),
+#'     data_sources = data_source(sales = sales)
+#'   )
+#'   commons_server("chat", agent)
+#' }
+#'
+#' shinyApp(ui, server)
+#' }
+#'
+#' @export
```
simonpcouch, line 2: Feels a bit awkward to me that `commons_app()` and `commons_server()` are documented together and this is on its own. I think either 1) `commons_app()` (simple, complete helper) and then 2) `commons_theme()` + `commons_server()` might feel better? Or all three in one?

pkg-r/inst/www/commons-chat/commons-chat.css
```diff
@@ -4,32 +4,30 @@
 
 /* ---- Chat-level polish ---------------------------------------------- */
 
-shiny-chat-container {
+/* These rules load before shinychat's own stylesheet (both arrive via the
+ * page theme), so overrides of shinychat classes need strictly higher
+ * specificity than shinychat's selectors to win regardless of order. */
+shiny-chat-container[id] {
```
gadenbuie: This looks like a problem we should fix in shinychat

pkg-r/R/theme.R
```diff
@@ -0,0 +1,51 @@
...
+    # User-message geometry, formerly a CSS override in commons-chat.css;
+    # shinychat exposes these as public theme variables.
+    "shiny-chat-user-message-border-radius" = "1.25rem",
+    "shiny-chat-user-message-padding" = "0.65rem 1.1rem",
+    ...,
+    preset = preset
+  )
+  bslib::bs_bundle(
+    theme,
+    sass::sass_layer(html_deps = list(commons_chat_dependency()))
```
gadenbuie, line 13: This dependency probably carries features that are beyond page style, right? If so, a user who provides their own theme would likely break other functionality in their commons app.

That points me toward maybe we'd still need a wrapper around the UI like `page_commons()` or `chat_commons()` -- or both -- even if they're thin wrappers that just inject the commons dependencies.

pkg-r/inst/app.R
```diff
@@ -83,9 +83,11 @@ greeting <- paste(
   "- <span class='suggestion'>Compare wetlands with the other habitats.</span>"
 )
 
-ui <- page_fillable(
-  title = "Biodiversity explorer",
-  commons_ui("chat", greeting = greeting)
+ui <- page_chat(
+  "Biodiversity explorer",
```
simonpcouch, line 91: Noting that this is bolded at the top of the page in the new default UI. For `commons_app()`, it is always "commons" that's bolded up there. (Fine that it's not able to be adjusted in `commons_app()`--we want to push people towards the 'real thing' sooner than later.)

> Image

Could `title` support an empty string? Or, in the cases that it's non-empty, could it not be bolded / be styled differently?

I think @skaltman found a nice feel (albeit with additional Posit branding) in the tiles agent, but as this currently is implemented, that bolded title makes the app feel more like "default Shiny" than it did before.

>  Image

simonpcouch (commented):
Nice! Broadly on board for the transition to `commons_theme()`. [EDIT: seeing that Garrick has some reservations there. Would defer to yall's concensus.]

I'm not big on the bolded, required page title. We can paper over this commons-side, but wanted to get this in front of yall first in case you're open to shinychat changes.

pkg-r/R/theme.R
cpsievert: Yea, it does currently bring in a JS dependency. I think we could/should get rid of it (https://github.com/posit-dev/commons/issues/173), but maybe long term there's a good reason to support it.

At least for now, I think we should just accept this as a limitation. If the user mistakenly uses `bs_theme()` instead of `commons_theme()`, it doesn't critically break things. If we do start pushing more into a JS layer, maybe we'd add a `commons_dependencies()` for this edge case situation.

Generally speaking, I would love to keep commons out of the business of wrapping shiny/shinychat's APIs (see also https://github.com/posit-dev/commons/issues/171). Besides the obvious (less to learn and maintain), I think it will also help to conceptually position commons within our AI stack (ellmer, shinychat, querychat, etc). More specifically, if commons could lean into the fact that it's just a supercharged ellmer client, then it will be more obvious how to do things like use a common client inside querychat, wrap an MCP server around it, etc.

pkg-r/inst/www/commons-chat/commons-chat.css
cpsievert: Yea, are you thinking we should `@layer` the styles?

pkg-r/inst/app.R
cpsievert: I think this is more of a question for @gadenbuie

pkg-r/R/theme.R
cpsievert: Agreed, but that said, would you be against the idea of `commons_app()` going away in favor of `shinychat::chat_app()`?

We would also need #171 for this to be a thing, but once we have that, I don't think we'd need `commons_app()`?

pkg-r/inst/app.R
gadenbuie: > Noting that this is bolded at the top of the page in the new default UI.

You can set the font weight via the theme with this Sass variable

```r
page_chat_theme(
  "shiny-chat-page-title-font-weight" = 300
)
```

or via the CSS variable

```css
:root {
    --shiny-chat-page-title-font-weight: normal;
}
```



> Could `title` support an empty string?

This is possible but requires `title = list()`. It's not generally a great idea, especially if you have more than one page, so the API is intentionally rough. I'd recommend against removing the title in general.

pkg-r/inst/www/commons-chat/commons-chat.css
gadenbuie: Not necessarily. I think we probably need to adjust our pattern around how we use global CSS variables so that these can be set on `:root`

pkg-r/R/theme.R
simonpcouch: I do think I'd be against that idea, yeah. I'd prefer that users have a one-liner they can test with without thinking about the shinychat implementation details.

cpsievert: It would still be a one-liner, though (`chat_app(commons_client)`)? In fact, that way `ellmer::live_browser(commons_client)` would also work (it's the same thing).

simonpcouch: The theming + provenance UI would not be applied in that `chat_app(commons_client)` case, right?

cpsievert: Ohh right, yea, I overlooked that, probably worth keeping just for that.

BTW, I would like for us to aim for not making provenance UI itself dependent on the theme existing. That is, a commons client + shinychat should be enough to emit the right markup/UI, and the theme just adds commons styling opinions that happen to diverge from vanilla shinychat.

pkg-r/inst/www/commons-chat/commons-chat.css
gadenbuie: Should be fixed by https://github.com/posit-dev/shinychat/pull/355

pkg-r/R/theme.R
cpsievert: Ok, updated the docs in [719f41f](https://github.com/posit-dev/commons/pull/169/commits/719f41f43cdab187feca87db0dd1150b73cb55c7)

pkg-r/inst/www/commons-chat/commons-chat.css
```diff
@@ -4,35 +4,33 @@
 
 /* ---- Chat-level polish ---------------------------------------------- */
 
-shiny-chat-container {
+/* shinychat proxies its public --shiny-chat-* variables through private
+ * --_* variables with the default in the var() fallback, so :root
+ * declarations here always win without specificity tricks. */
```
gadenbuie:
```suggestion
/* Set shinychat CSS variables */
```

cpsievert: Thank you!

pkg-r/inst/app.R
cpsievert: @simonpcouch if you prefer, we can have `commons_theme()` default to that font weight. Just lmk

cpsievert:
@simonpcouch should be ready for another look.

simonpcouch (approved):
Thank you! 

Just bumped that font weight here.

Let's squash and merge. :) I can send some PRs out to early-adopting packages once this is in.

pkg-r/R/theme.R
```diff
@@ -0,0 +1,17 @@
+#' @rdname commons_server
+#' @export
+commons_theme <- function(..., preset = "shiny") {
+  theme <- shinychat::page_chat_theme(
+    "shiny-chat-page-title-font-weight" = 300,
+    # User-message geometry, formerly a CSS override in commons-chat.css;
+    # shinychat exposes these as public theme variables.
```
cpsievert:
```suggestion
```

PR #172, opened 2026-08-25
Remove client-side scroll-jump logic now that shinychat fixes it upstream

cpsievert:
## Summary

Removes the client-side scroll-jump logic in `commons-chat.js`. It worked around shinychat's stick-to-bottom scroll chasing the height change when a tool card expanded/collapsed, by synthesizing wheel events to trigger the library's escape hatch.

This logic is no longer needed: shinychat now calls that escape hatch (`stopScroll()`) directly whenever a tool card, tool group row, or long tool-call preview is toggled (posit-dev/shinychat#348, merged).


simonpcouch (approved):
ty!!

PR #190, opened 2026-08-27
standardize user-facing trusted calculation language

### LINKED ISSUE #186: Standardize user-facing language around trust/governance
The activity labels use "trusted calculation," but the hover-over text on the provenance marker uses "governed calculation." 

Image

I think "trusted calculation" is probably more understandable to users, but "governed" might be more technically accurate.

simonpcouch:
Closes #186. My preference here was for trusted rather than governed. Both of these words have technical meanings (that sometimes vary by context, admittedly) that don't quite fit here, but I think trusted has a more straightforward colloquial meaning. This might especially apply for the measure (i.e. extracted from a Quarto report or Shiny app) case, as in "I wrote this so I trust it."


skaltman (approved):
I agree, I think "trusted calculation" is better.

PR #197, opened 2026-08-27
Replay the live chat's provenance marker in review transcripts

### LINKED ISSUE #173: Eliminate the JS layer
After #172, it seems the only logic left for JS is related to provenance tooltips. I'm not entirely sure if this is still needed at this point. And if it is, It smells like something that could be provided by shinychat or Bootstrap.

cpsievert:
Closes #173 

The trajectory review now replays the exact provenance marker the user saw
in the live chat — same `<shiny-aside>`, same popover — instead of rendering
a custom pill with ~140 lines of hand-maintained tooltip CSS/JS, which this
PR deletes.

- `add_message_provenance()` attaches `provenance_aside(tag, include_cited = TRUE)`
  to each exchange's final answer as a trailing markdown chunk — the same
  shape a live-streamed aside arrives in — so shinychat's own aside grouping
  and floating-ui popover handle rendering and positioning.
- `include_cited` (default `FALSE`, live behavior unchanged) keeps a visible
  marker for Cited answers in the review, which the live chat deliberately
  leaves bare.
- The question list keeps its pill — it's server-rendered HTML outside
  shinychat's React tree, where `<shiny-aside>` would be inert — now with a
  `bslib::tooltip()` (parent commit).

## Transcript rendering: replay, not static serialization

The transcript no longer goes through static `chat_ui(messages = ...)`. That
path folds mixed content (tool cards plus markdown text) into a single
raw-HTML island, escaping the text and hiding it from aside grouping —
contrary to shinychat's own docs ("strings are still interpreted as markdown
as long as they're not inside HTML").

Instead, `replay_transcript()` mirrors shinychat's own bookmark-restore
idiom (`client_set_ui()`): each message's contents are yielded from a coro
generator through `chat_append()` once the chat element is bound, so every
chunk is classified markdown vs HTML exactly as in the live stream. Tool
cards render as shinychat's native tool groups, htmlwidget dependencies are
collected per chunk, and thinking blocks keep their styling — none of which
held under static serialization. Exchange decorations are seeded after the
replay in the same flush, so message indices line up.

A static-`chat_ui()` fix upstream in shinychat would still be welcome, but
this PR no longer depends on or works around that path.

## Validation

`test-trajectory-review.R`, `test-provenance.R`, and `test-chat.R` pass,
including coverage that transcripts carry the aside marker as a trailing
markdown chunk for A and B outcomes. Full suite left to CI.

Verified manually in a local review app with fabricated trajectories
covering every provenance outcome plus a `run_sql` tool call (example app
kept local, not in this PR):

- All three aside markers render as shinychat's native grouped pills, with
  working popovers (opaque background; the transparency issue seen with the
  old hand-rolled CSS is gone).
- The tool call renders as shinychat's native tool group in the same
  message flow as the aside markers.
- Exchange selection syncs in both directions: transcript click updates the
  review pane and sidebar highlight, and sidebar clicks highlight the
  corresponding transcript messages.

Browser/live-warehouse tests not run.


simonpcouch (approved):
Thank you!

PR #199, opened 2026-08-28
override `...` pending indicator with a shimmering "Working..."

simonpcouch:
Hooks into https://github.com/posit-dev/shinychat/pull/363.


simonpcouch:
Supersedes #144.

Issue #171, opened 2026-08-25
Consider removing `commons_server()` from API


cpsievert:
And using `shinychat::chat_server()` instead.

Currently, `commons_server()` does 3 things:

1. Creates a `commons_server_start` span. 
    *  Could be removed considering that Shiny already has instrumentation for the session's lifecycle?
2. Creates a conversation id. 
    * This is a hack (that I added) and can go away when we have https://github.com/posit-dev/shinychat/pull/343
3. "Pre-warms" the client (i.e., builds the index, etc).
    * We might want to revisit how this works, but regardless, it might be better if this was explicit rather than implicit. Essentially the (recommended) API would change from:

```R
client <- ...
commons_server("chat", client)
```

to

```R
client <- ...
client$prewarm()
chat_server("chat", client)
```

I'm not sure if it's possible today, but at least at some point it should be possible to prewarm/cache the index independently of the shiny runtime. In that case, you might not want the pre-warming at all?

cpsievert:
For more context, a motivator for this change is for things like `querychat` which calls `shinychat::chat_server()`. If `commons` has it's own `commons_server()`, it's harder to imagine how `querychat` would support a commons layer. Obviously, there are other compatibility concerns we'd have to address to make that happen, but I also think it helps commons position itself conceptually in the landscape if it's essentially just an `ellmer` agent.

PR #202, opened 2026-08-28
Detect history restores in the agent instead of a shinychat hook

cpsievert:
Towards #171

## Summary

`commons_server()` registered an `on_restore` hook with shinychat solely to queue the restore reminder (the hidden note telling the model that R state behind a restored transcript is gone). But every shinychat history operation that brings in foreign turns — page-load restore, in-session switch, bookmark restore — funnels through `client$set_turns()`, which the commons agent already overrides. The agent can observe restores itself; no shinychat hook is needed.

`set_turns()` now queues the reminder when a non-empty history arrives that isn't a content-prefix of the current turns (i.e. it came from outside this session), and `commons_server()` drops its last history hook — leaving it as just prewarm plus a setup span, a step toward `shinychat::chat_server()` as the front door (#171).

## Details

- Comparison is content-based (role + text via `turn_signature()`), because shinychat replays stored turns through `ellmer::contents_replay()` and object identity is meaningless.
- Truncations (edit/branch nav to a prefix) and redundant re-restores of the same history do **not** queue the reminder; a pending reminder survives them and is only consumed by a completed turn or a cleared history.
- The trigger is slightly broader than before: any programmatic `set_turns()` of foreign history queues the reminder, not just shinychat restores. This is arguably more correct — the reminder's condition ("this transcript's R state isn't real") holds regardless of who loaded the turns.
- With the hook gone, the public `agent$queue_restore_reminder()` method had no remaining callers and is removed; the reminder is now entirely agent-internal.

## Testing

- The testServer-based hook test is replaced by agent-level tests covering truncation, redundant restore, foreign history, and clearing.
- Full suite matches the `main` baseline (remaining failures are pre-existing in this environment: run_r worker tests and the dev-ellmer citation-scan interactions).


simonpcouch (commented):
Coming back to this after revisiting #220. I agree with you (and Fable😄) that `commons_theme()` + `commons_server()` now seems like the right set of abstractions. Given that, are you still hoping to see these changes go in? Happy to work from this, but relying on the public shinychat hook feels like the more principled approach to me ATM

cpsievert:
Yeah, assuming we keep `commons_server()`, I'm happy to close this one

PR #218, opened 2026-08-28
Revamp tool preview UI

### LINKED ISSUE #216: Revamp tool UI for trusted calculations
Currently, measure and metric tool UI is pretty limited.

They look like e.g.:

> Image

> Image

I think we could do much better here. The name of the measure/metric and arguments are probably not very informative for users. For one, for measures and metrics, is there some limited documentation derived from the metadata or data-dict that would be useful for end-users to see?

As for "can be shown to be correct", I don't think it's worth trying to encapsulate measure source or metric SQL into a single thing that could be made visible. Is it somehow possible to link directly to the source, whether that's the provenance link or the place in the source code for the agent itself?

The same could probably go for the calculations that are automatically derived from Snowflake/Databricks, though I think we should honestly leave that for later.

In the `search_pool()` results, what if we showed the rendered markdown rather than the markdown itself?

> Image

simonpcouch:
Closes #216. More user-friendly presentation of the trusted calculations and context.

https://github.com/user-attachments/assets/f23a2923-b95c-4adb-a743-b0b53e761b24

### DIFF
--- a/pkg-r/R/measures.R
+++ b/pkg-r/R/measures.R
@@ -269,7 +269,11 @@ is_measure_list <- function(x) {
   is.list(x) && !inherits(x, "ellmer::ToolDef")
 }
 
-measure_schema_text <- function(td, source_names = character()) {
+measure_schema_text <- function(
+  td,
+  source_names = character(),
+  heading = tool_name(td)
+) {
   props <- tool_properties(td)
   args <- if (length(props) > 0) {
     paste0(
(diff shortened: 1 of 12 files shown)


pkg-r/R/measures.R
```diff
@@ -18,6 +18,7 @@
 #' `string`, `integer`, `number`, `boolean`, `enum[value, ...]`, or an array
 #' such as `string[]`. Without a declaration, commons infers the type from the
 #' default, falling back to `string`.
+#' URL-valued `@provenance` tags appear as source links in the measure result.
```
simonpcouch:
```suggestion
```

PR #219, opened 2026-08-28
Explain the badges in a modal

### LINKED ISSUE #217: Document what the provenance badges mean in the deployed app
On hover/click over the trust badges, we could add a small `(i)` information button in the BRHS that brings up a popover describing the trust system.

Stubbing that out:

```
This application works by asking an AI agent to use calculations vetted by humans when possible. When it can't find a relevant calculation, it will try to write code itself to answer the question. Based on the agent's path, the application will assign a tag to the response:

* Trust icon: The agent used only a vetted calculation in answering the question.
* Cited icon: 
* Untrusted icon: 
* No icon:
```

simonpcouch:
Closes #217. Just a proof of concept.

https://github.com/user-attachments/assets/26f8095c-3b8c-4707-9784-2a31d253b670


pkg-r/inst/www/commons-chat/commons-chat.js
```diff
@@ -0,0 +1,31 @@
+(() => {
+  const triggerSelector = ".commons-provenance-info-trigger";
+
+  const openModal = (trigger) => {
+    const template = trigger.nextElementSibling.firstElementChild;
+    const modal = template.cloneNode(true);
+    document.body.append(modal);
+
+    const instance = new window.bootstrap.Modal(modal);
```
cpsievert: This could all be done without custom CSS and JS via `shiny::actionLink()` + `shiny::showModal()`.

Might also be worth considering using `bslib::offcanvas()` so users can better see the chat and the "about"

simonpcouch: Thanks, giving this a go!

simonpcouch: Okay, right—the simplest `shiny::actionLink()` + `shiny::showModal()` doesn't quite work because the icon lives inside of the `shiny-aside`. Just pushed a take on this that includes a bit of JS so that the button is still clickable. (Notably, this would be another responsibility of `commons_server()`.)

simonpcouch:
Oh, another point from Sara:

> I think we should also give the model instructions about how to answer if the user asks about the trust tags. I just tried asking and it gave an answer that would probably not be helpful to the average user

A bit confusing as well as the model actually doesn't see which answers were tagged as what.

simonpcouch:
Okay, here's a draft of the modal design:

> Screenshot 2026-08-28 at 3 35 53 PM

skaltman:
Having this will be really helpful.

I would maybe avoid "human-vetted calculation," in case people think that means each one is reviewed in some way.

Maybe something like, "...a trusted calculation that was developed and maintained by the agent authors."

PR #221, opened 2026-08-28
Remove commons_server() in favor of shinychat::chat_server()

cpsievert:
Stacked on #202.

With shinychat now managing conversation identity and history restores first-class, `commons_server()` no longer adds much on top of `shinychat::chat_server()`, so this PR removes it. Custom apps now compose `shinychat::chat_server()` directly, and the idle-time pre-warm it used to do is exposed as an exported `commons_prewarm()` helper:

```r
# before
commons_server("chat", agent)

# after
commons_prewarm(agent)
shinychat::chat_server("chat", client = agent)
```

The two call sites differ in error handling: a direct `agent$prewarm()` — typically cache warming ahead of deployment — lets failures propagate, so a cold cache fails the deploy; `commons_prewarm()` downgrades failures to warnings, since pre-warming is a pure optimization (everything it builds is rebuilt lazily at first use) and an error escaping a `later::later()` callback would stop the whole Shiny app.

## Testing

- New tests: `prewarm()` propagates failures; `commons_prewarm()` downgrades a prewarm failure to a warning (driven through `later::run_now()`); `commons_app()` prewarms the agent on idle; a commons agent runs under `shinychat::chat_server()` via `shiny::testServer()`.
- Deleted: the `commons_server()` tests.


cpsievert:
@simonpcouch I think I'd also be open to just keeping `common_server()` for the easy "batteries included" API, but I generally still think we should go in this direction where "batteries become optional" and it's easy to pick and choose via a composable, functional, API

simonpcouch:
Closing for now, open to opening up more composable entry points in the future.

PR #224, opened 2026-08-28
Idea: provenance info explainer as a web component

cpsievert:
Stacked on top of #219 — offering this as a rough direction before signing off, not a finished PR.

## The idea

Instead of an `actionLink` + server observer + client-side re-binding, the aside markup just includes a `<commons-provenance-info>` web component. The component renders its own trigger button and opens a Bootstrap offcanvas panel (sliding side panel rather than a modal) explaining the outcomes. All of your latest copy tweaks are carried over verbatim.

## Why this direction

- **No binding lifecycle to manage.** shinychat mounts aside HTML after Shiny's input-binding pass; a custom element upgrades whenever it connects, so there's nothing to observe or re-bind (the MutationObserver goes away).
- **No server code at all.** Nothing to wire into each app, no input id to keep in sync, nothing that breaks inside modules.
- **Side panel instead of modal.** It doesn't cover the answer being explained, and Escape/backdrop/focus handling come from Bootstrap for free.

## Honest caveats

- Unit and browser tests pass, but I haven't actually eyeballed this in the app — no dark-mode check, no look-and-feel pass. Treat it as a sketch of the direction, not finished work.
- The explainer copy now lives in `commons-chat.js` rather than R, so future text tweaks happen there.
- One thing to decide: your last commit relabels the first outcome "Trusted" in the explainer, but the marker itself still says "Verified answer" (that label is shared with the Python package via the fixture). Worth aligning one way or the other.

Feel free to take this over, cherry-pick from it, or close it if you'd rather keep the modal approach.


simonpcouch:
> Screenshot 2026-08-28 at 3 38 27 PM

cpsievert:
FWIW, I think you could basically take this same approach with a modal if you prefer to. Off canvas is kind of nice though, 'cause you can get rid of the gray background and position it b/l/t/r.

simonpcouch:
Ah, cool! Just poked at this for a bit and, in 217, adopted your approach with the modal. Will sit with this for a bit.

simonpcouch:
I ended up adopting this approach in the modal! Thanks for demoing this out!

PR #229, opened 2026-08-31
Publish the Python package as `commons` on PyPI

jat255:
We own the `commons` name on PyPI now, so the Python package is published under it instead of `posit-commons`. The import name was already `commons` and does not change.

## Summary

The distribution name in `pkg-py/pyproject.toml` becomes `commons`, and the lockfile, both READMEs, the package docstring, and the comment in `py-release.yaml` follow.

Because the distribution and import names are now the same, two things that existed only to support the mismatch are removed rather than updated. Hatchling derives the package directory from the distribution name, so `[tool.hatch.build.targets.wheel] packages` was needed to point it at `src/commons`; it now infers the same directory unaided. And `pkg-py/tests/test_packaging.py` was written to make the split hard to reverse by accident, including a test asserting that `posit_commons` is not importable, which no longer pins anything. What remains pins the distribution name, the import name, and the `py.typed` marker.

## Review notes

The trusted publisher on the `commons` PyPI project is already configured against this same workflow file, so releases keep working over OIDC with no repository change beyond that comment.

This breaks `pip install posit-commons`, deliberately and with no deprecation path. `posit-commons` only ever held one placeholder release (0.1.0.dev0), nothing depends on it, and the repository's own guidance is that neither package has been publicly released. The intent is to delete that project from PyPI rather than yank it or keep it as a redirect.

Two consequences worth weighing, neither visible in the diff:

- Deleting a PyPI project frees its name for anyone to re-register, because PyPI only blocks reuse of deleted versions within a surviving project, not deleted project names. Someone could publish a hostile `posit-commons` that stale links would install.
- The `commons` project currently lists no files at all, which is consistent with its earlier releases having been deleted. PyPI permanently blocks reuse of deleted version numbers, so the version here is 0.1.0.dev1 rather than 0.1.0.dev0. That number has never been published under either name, so tagging this after merge exercises the release workflow end to end instead of failing on a duplicate.

## Testing

`uv sync --locked --all-extras`, `uv run ruff check`, `uv run ruff format --check`, `uv run pyrefly check`, and `uv run pytest` all pass, which is what `py-check.yaml` runs. The `--locked` run is the one that matters most here, since it confirms the regenerated lockfile matches the renamed project.

Beyond the standard checks, both the wheel and the sdist were built and inspected, confirming that dropping the hatch packages setting still yields `commons/` with `py.typed` at the root. The release workflow's tag check was also simulated against the new wheel filename, confirming it accepts `py-v0.1.0.dev1` and rejects a mismatched tag. That check parses the version out of the filename, which the distribution name is part of, so the rename could have broken it.

No tests were added. `test_packaging.py` shrank because the behaviour it guarded no longer exists.

### DIFF
--- a/.github/workflows/py-release.yaml
+++ b/.github/workflows/py-release.yaml
@@ -1,7 +1,7 @@
 name: py-release
 
 # Publishes pkg-py to PyPI via trusted publishing (OIDC); no API token is stored.
-# PyPI's publisher config for posit-commons names this file, so renaming it
+# PyPI's publisher config for commons names this file, so renaming it
 # breaks releases with an OIDC error until PyPI is updated to match.
 on:
   push:
--- a/README.md
+++ b/README.md
@@ -32,10 +32,13 @@ pak::pak("posit-dev/commons/pkg-r")
 
 To learn more, see `vignette("commons", package = "commons")`.
 
-To install the Python package from PyPI, run:
+To install the [Python package from PyPI](https://pypi.org/project/commons/), run:
 
 ``` sh
 pip install commons
+
+# or, using uv (https://docs.astral.sh/uv/):
+uv pip install commons
 ```
 
 Then, import with:
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ To learn more, see `vignette("commons", package = "commons")`.
 To install the Python package from PyPI, run:
 
 ``` sh
-pip install posit-commons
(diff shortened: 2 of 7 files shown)


jat255:
@simonpcouch as an FYI, this PR switches us from `posit-commons` to just `commons` after a gracious gesture from the owner of that name on PyPI

PR #180, opened 2026-08-27
Render provenance and citation markers as grouped icon asides

cpsievert:
Supersedes #166. Closes #148 (default shinychat citation UI) and closes
#149 (minimal markers) — the look #149 asked for, through the mechanism
#148 asked for.

## Where things stand

Both PRs chase the same destination: small icon markers instead of
pills, with a one-shot pulse on mount. #166 gets there with commons-owned
CSS that restyles shinychat's compact-aside DOM state by state; this PR
gets there by changing what commons *asks shinychat to render* — labeled
identity asides — and confining commons CSS to shinychat's public
`--shiny-chat-aside-marker-*` theming variables.

How they compare, at a glance:

| | #166 | This PR |
|---|---|---|
| Marker look | Small icon, pulse on mount | Same look (warning icon identical to #166's; citation quote mark redrawn as a fixed path) |
| Mechanism | CSS overrides of shinychat's compact-aside DOM | shinychat's aside grouping + public theme variables |
| Multiple citations in one paragraph | Flattened to one dot, indistinguishable from a single citation | One marker with a "+N" overflow count and a carousel popover; every citation stays reachable |
| Hover/focus/expanded states | Re-implemented per tone in commons CSS | Inherited from shinychat |
| Cited answers in review transcript | Marker removed | Unchanged here; see note below |

## What this PR does

Conceptually, three moves:

- **Citations render as labeled identity asides** instead of compact
  numbered chips, so shinychat owns the marker entirely — grouping,
  overflow counts, popovers, and interaction states.
- **All three markers (Verified, Untrusted, citation) render
  icon-only.** The label stays in the DOM for the pill's `aria-label`,
  and commons CSS touches only shinychat's public
  `--shiny-chat-aside-marker-*` variables.
- **Singleton citation popovers gain a body title** (icon + source
  label), since shinychat only renders its own title row for grouped
  asides.

A local example app renders every outcome — including a three-citation
paragraph and a singleton citation — for manual comparison against the
review transcript. It's not part of the package; source below.

<details>
<summary>Example app: all provenance outcomes and grouped citations</summary>

```r
# Save as app.R and run with shiny::runApp().
#
# What a live commons chat renders for each provenance outcome:
# Verified and Untrusted answers end with a marker; Cited answers show
# their citation asides instead — including a paragraph with several
# citations, which shinychat accumulates into one marker with a "+N"
# overflow and a carousel popover. A fourth message carries a lone
# citation: shinychat only renders its popover title row for grouped
[...]

PR #230, opened 2026-08-31
Python citation block parser + shared `citations.json` fixture

jat255:
## Summary

Adds `parse_commons_citation()` to the Python package. The citation grammar and the `commons.citation.candidates` record move into a shared fixture (`tests/shared/citations.json`) that both suites read.

**R changes are tests only — no functional R code changed.**

## What changed

- Python: new `parse_commons_citation()` and a frozen, self-validating `CitationDecision` dataclass. Status/kind vocabularies come from `Literal` aliases via `get_args()`; `label`/`kind` exist only for accepted decisions, and only malformed decisions lack a quote.
- Shared: parser test cases and the candidates-record shape promoted to `tests/shared/citations.json`.
- R: 
  - `test-citation-scan.R` now reads parser cases from the fixture via `shared_fixture()`. 
  - Three hand-written tests were moved to the shared fixture with identical bahavior
  - Seven new cases added: One new test checks the JSON structure of a citation-decision record. These records are written to trace spans and read back by `trajectory_read()`; no test previously verified the two sides agree on the field names, so a mismatch would have failed silently.

## Citation block format

The shared test fixtures enforce the following behavior:

> A citation consists of exactly one contiguous run of > lines holding the verbatim evidence; all other lines are explanation and never verified. Zero or two runs is malformed. Two details worth noting:
>
>  - A wrapped quote parses: the continuation line becomes explanation, never verified.
> - The marker is ^> ? with no indentation allowance; an indented > line makes the body malformed.

## Known divergence between R and Python

Explanation trimming uses each language's whitespace class, so a trailing U+00A0 survives in R but not Python. Explanation is never verified, so this is cosmetic; it's documented in the fixture.

## Reviewer notes (for R)

- The new decision test drives all three producers (accepted/rejected/malformed) from the fixture rather than a hand-built list — more indirection than usual, but the point is that the real producers (for both R and Python) agree with the recorded shape. I'm open to feedback on whether those three statuses fully describe what `citation_scanner()` records.
- A reminder that any files under `tests/testthat/fixtures/shared/` are generated by `scripts/sync-shared-fixtures.sh`, so they don't need to be separately reviewed.

## Testing

- Python: 68 tests, ruff, pyrefly — all pass.
- R: `test-citation-scan.R` passes in full (25 tests, 217 expectations) against the rebased tree. `R CMD check` not run locally (missing R deps); left to CI.


jat255:
## Summary of the R changes

This is for the reviewer of the R code. No R code changed in this PR. The changes are in `tests/testthat/test-citation-scan.R`, plus the generated fixture copy.

### What changed and why

`parse_commons_citation()` now has a second implementation in the Python package, so its grammar became a cross-language contract. Under the rule in `tests/shared/README.md`, that contract belongs in one fixture that both suites read, not in a hand-written test on each side. The parser cases moved into `tests/shared/citations.json` and the R suite reads them through the existing `shared_fixture()` helper.

A second test is new coverage rather than a move. It pins the shape of one entry in `commons.citation.candidates`. That JSON is what `trajectory_read()` parses back out of the span, and nothing asserted its field names before.

### Behaviour delta

None. `R/citation-scan.R` and `R/citations.R` are untouched, so every input produces exactly what it produced before.

### Deleted tests and what replaced them

Three hand-written tests were deleted:

- `parse_commons_citation splits explanation and blockquote`
- `parse_commons_citation is NULL without exactly one blockquote`
- `a bare '>' line continues the quote run`

All three cases are in the fixture, with the same inputs and the same expectations. Seven more cases came with them:

- a marker with no following space
- explanation on both sides of the evidence
- a wrapped line that stays explanation
- a body that is only evidence
- an indented marker
- an empty body
- a body with no blockquote

### Blast radius

`parse_commons_citation()` has one caller, `close_citation()` inside `citation_scanner()`. That path is unchanged and its own tests are unchanged. The new decision test calls `render_citation_aside()` and `project_citation_text()` but only reads the returned decision, never the HTML, so the grouped icon aside markup from #208 and #180 does not reach it.

### Why it is safe

The full `test-citation-scan.R` file passes on this branch after the rebase onto current main. That is 25 tests and 221 assertions, run against the real `R/citations.R` and `R/citation-scan.R`.

The expectations in the fixture were written first and then confirmed against the running R parser, not read out of it. Two of them contradicted the Python plan for this task, and R won both. A wrapped quote parses, with the continuation line becoming explanation. An indented `>` line is not evidence.

`R CMD check` and `devtools::test()` did not run. This machine has no `ellmer`, `duckdb`, `coro`, or `shinychat` installed, so `pkgload::load_all()` fails on imports. The file was run instead by sourcing the three R files it needs into a testthat session. As a control, `test-citations.R` was run the same way on this branch and on main. It fails 12 of 21 tests identically on both, all of them from the missing packages, and this PR does not touch that file.

### What deserves scrutiny

The decision test drives three different producers through a `switch()` on the status in the fixture, which is more indirection than a normal testthat test. The choice was deliberate. Asserting on a hand-built list would pin nothing, because the point is that the real producers agree with the recorded shape.

Worth a second opinion: whether `commons.citation.candidates` is fully described by these three cases. Accepted, rejected, and malformed are all that `citation_scanner()` records today.

Files under `tests/testthat/fixtures/shared/` are generated by `scripts/sync-shared-fixtures.sh`. Skip them.

jat255:
@simonpcouch this includes a minor change to R citation tests since they are using a the shared fixture with the Python side. It also added a few new tests to ensure citation formats match in both implementations. There's no functional changes in runtime R code, so it's all additive, but just an FYI.

simonpcouch (approved):

PR #231, opened 2026-08-31
Persistent, content-addressed context store + idle-time prewarming

cpsievert:
## What changes

Building the ragnar context index (DuckDB + FTS) was previously the most expensive part of constructing a `commons()` agent, and every session paid for it: the store lived in memory, scoped to the layer object, so a new process meant a full rebuild. This PR makes the store a **persistent, content-addressed file on disk** and adds explicit **prewarming hooks** so the cost can be moved off the critical path — or off the deployed app entirely.

## The key ideas

**Content-addressed, not session-addressed.** The store's filename hashes
the layer's documents plus the ragnar/duckdb package versions (whose on-disk format the store depends on). Correct invalidation falls out for free: edit a doc or upgrade a dependency and you get a new file; change nothing and every session, process, and deployment opens the same one. There's no cache key management to get wrong because the content *is* the key.

**Build once, then everyone just opens a file.** Builds go to a temp file in
the same directory and are renamed into place atomically, so no reader ever observes a partial store, and two concurrent builders degrade to "loser discards an equivalent store." Readers connect read-only (`ragnar_store_connect()`'s default), so concurrent Shiny sessions across processes can share one store file safely. The cache is size-capped (256 MB by default) and pruned LRU, with an mtime touch on open keeping actively used stores young. Temp files from interrupted builds (killed processes, crashes) are reaped once they're a day old, so debris can't accumulate outside the size cap.

**The cache root is deployment-aware.** Resolution order: an explicit
option/env override, then Connect's persistent content data dir (`CONNECT_CONTENT_DATA_DIR`, survives redeploys where enabled), then an `app_cache/commons/` directory beside the app (the sass convention: per-app scratch on hosted platforms, excluded from deployed bundles), then the per-user cache dir. On ephemeral filesystems (Connect Cloud resets disk to the bundle), the store simply rebuilds once per cache lifetime instead of once per process. And because the store is just a file, you can **ship a warm cache with the app**: point `commons.context_cache` at a directory inside the app and run `prewarm_context()` before deploying. An unwritable cache root warns once and falls back to a per-session tempdir — caching can never take down the app.

**Prewarming is explicit and failure-semantics-aware.** `agent$prewarm()`
splits into `prewarm_context()` (build the index) and `prewarm_sources()` (background pin download). A direct call lets failures propagate — warming ahead of a deploy should fail the deploy. `commons_prewarm()` is the Shiny path: it defers warming to a `later::later()` callback and downgrades failures to warnings, since an escaping error would stop the app and prewarming is a pure optimization. Outside a running Shiny app (e.g. a pre-deploy script) there's no event loop, so it warms synchronously.

## How this relates to the old design

The previous `commons_context_layer` was a plain list carrying a `cache` environment, and `context_store()` memoized an **in-memory** (`:memory:`) ragnar store in `layer$cache$store` on first search. The environment was the whole trick: reference semantics meant every copy of the layer shared one store, and because `augment_context_layer()` built a fresh object (with a fresh, empty environment) whenever docs changed, a stale store could never be reused. In effect it was lazy memoization keyed by **object identity** — which meant the store died with the R process, and "sharing" only worked if you literally passed the same object around. Every new session, Connect worker, or redeploy rebuilt the index from scratch.

This PR keeps that design's shape — lazy build, cached on the layer, fresh store when docs change — but changes what the cache *is*:

- **Object-keyed → content-keyed.** "Same object → same store" becomes
"same docs + same ragnar/duckdb versions → same file on disk." Sharing no longer requires passing one object around; independently constructed layers with identical docs hit the same cache entry, across processes and deployments.
- **`:memory:` → persistent file.** The expensive build happens once per
content version per cache root instead of once per session.
- **The environment survives, demoted.** The R6 `ContextLayer`'s private
`store` field plays the old environment's in-process role (repeat searches in a session don't reopen the file); the disk cache sits underneath it as the cross-process layer. The fresh-store-on-doc-change property is preserved twice over: augmenting docs still creates a new layer object, *and* different docs hash to a different filename anyway.

The old code paths are fully removed — `new_context_layer()` no longer creates the list-plus-environment structure, and nothing references `layer$cache` anymore.

## What this means in practice

- First search in a warm-cache session: milliseconds (open a file) instead
of a full chunk + index build.
- `commons_server()` prewarms automatically during post-startup idle time,
while the user reads the welcome message.
- Deploy-time warming is a first-class workflow, not a hack.

## Prior art

Much of the logic and motivation here is adapted from the **sass** R package, whose file cache (`sass_cache_*()`) has run in production Shiny deployments for years. The pieces we borrow directly: the cache-root resolution ladder (explicit option → env var → `app_cache/` beside the app on hosted platforms → per-user cache dir), content-addressed keys so invalidation is structural rather than managed, size-capped LRU eviction, and the posture that caching is a pure optimization that must never take down the app. sass's cache has survived every hosting quirk Connect and Shiny Server have thrown at it; following its precedent is deliberate risk reduction, not novelty.

## Escape hatches

- `options(commons.context_cache = FALSE)` (or `COMMONS_CONTEXT_CACHE=false`)
disables persistence — stores build in memory per layer, as before.
- `options(commons.context_cache_max_size)` caps total cache size (default
256 MB) with least-recently-used eviction — the single pruning knob. Following cachem's precedent, a single store larger than the cap is kept (with a one-time warning) rather than evicted into a rebuild loop.

## Also in this PR

- Layer objects (`commons_data_source`, `commons_semantic_layer`,
`commons_context_layer`) become R6 classes with private state, so connections, processes, and caches have explicit ownership.
- Metric tool results now carry title/description metadata for richer
display.
- Review-driven hardening: the cache option is validated (a non-string,
non-`FALSE` value aborts rather than creating a directory named `"TRUE"`), and telemetry span attributes can never abort prewarming. A store that fails to open (e.g. unlinked by a concurrent pruner between the existence check and the connect) warns — with a Shiny notification when running in an app — and is rebuilt once; a second failure still surfaces.


simonpcouch (commented):
Could you do a pass on the code comments (incl the roxygen) and also write a couple sentences in your own words about what this PR does and what I should pay attention to? 

On first glance, it seems like this exports a new function in the namespace instead of making slots on the data source / context layer public. Open to this idea, but any reason why? It also seems like we've introduced a number of user-facing envvars and options. Should any be function arguments?

cpsievert:
> in your own words about what this PR does?

The core idea is a persistent cache for the ragnar store, which should help with loading times, especially for larger stores. With the current in-memory approach, the cache is lost every time the server goes to sleep (i.e., after ~10 mins of inactivity), and does nothing for deployments that dedicate one process per visitor. Instead, this PR takes a file-based cache approach (works since Ragnar stores are backed by DuckDB files), which allows the cache to persist even across deploys is some situations. As a result, this should reduce the need for workarounds like https://github.com/posit-dev/tiles-agent/pull/37 to improve startup time.

> what I should pay attention to?

Much of the implementation here is borrowed from `sass::sass_file_cache()`, which has battle tested code for caching of Sass -> CSS compilation (needed for bslib apps), so I wouldn't worry too much about the lower level implementation details of how things like pruning work. The main thing would be signing off on the idea that file-based caching is sufficient for caching knowledge stores, what the user experience is like, and how we document it. You may also just want to double check how we compute the cache key (`rlang::hash()` of the store input).  

> this exports a new function in the namespace instead of making slots on the data source / context layer public

I believe this is in reference to `commons_prewarm()`? The main idea here is to provide a way to build the cache offline and deploy it with the app. This way you could (in a sense) guarantee that even the first visit leads to a cache hit. Since you've looked at these changes, I've reworked `commons_prewarm()` to make this more obvious (see 672fd59).

>  It also seems like we've introduced a number of user-facing envvars and options. Should any be function arguments?

The options this PR adds is `commons.context_cache` and `commons.context_cache_max_size`. Largely speaking, most users won't need to know about them since the defaults are designed to do the right thing on our hosted platforms. However, if a user has a reason to use `commons_prewarm()`, they'll become relevant, so I think it makes sense to just document them there and avoid another public function (at least for now).

simonpcouch (approved):
Thanks so much for writing this up. Sheesh, I've found myself so conflicted on this one, but I think we should take it. Currently, tiles-agent indexes locally in 1.37 seconds. This is long enough to be worth addressing on its own, and we expect this is a relatively small app compared to other use cases we have in mind. So, definitely needs doing.

From my understanding, both `commons_prewarm()` and `agent$prewarm()` allow the user to build the context cache synchronously before deployment. Beyond `$prewarm()`, the wrapper requires `cache_dir`, converts failures into warnings, and reports where the cache was created. It doesn't _also_ configure the option/envvar for the user; for deployment, the user would still have to set the option or envvar to note where the cache is.

As-is, this PR had:

```r
# agent.R
library(commons)

options(commons.context_cache = "commons-cache")

make_agent <- function() {
  commons(
    client = ellmer::chat("anthropic/claude-sonnet-5"),
    data_sources = data_source(cars = datasets::mtcars),
    context_layer = context_layer("context/context.md")
  )
}
```

Then, in `deploy.R`:

```r
# deploy.R
source("agent.R")

commons_prewarm(
  make_agent(),
  cache_dir = "commons-cache"
)

rsconnect::deployApp()
```

I just sent up some changes that propose removing the wrapper in favor of (optionally) using `$prewarm()` directly:

```r
# deploy.R
source("agent.R")

make_agent()$prewarm()

rsconnect::deployApp()
```

If you're game for those changes, I'd say let's bring this in!

cpsievert:
> Sheesh, I've found myself so conflicted on this one, but I think we should take it.

In case it helps motivate further, it's worth nothing that `prewarm_on_idle()` blocks Shiny from processing the _next_ flush, meaning things like history restoration, dynamic UI, etc., all need to wait for prewarming to finish.

> If you're game for those changes, I'd say let's bring this in!

LGTM 👍

PR #233, opened 2026-09-01
Python streaming citation scanner + shared projection fixtures

jat255:
## Summary 

Adds the citation streaming scanner (implemented on the R side as `citation_scanner()` in `pkg-r/R/citation-scan.R`) that rewrites a verified `<commons-citation>` into a `<shiny-aside>` and drops every other reserved element, so model-authored markup cannot reach the browser.

**R side changes are tests only.**

- Python: `_citation_scan.py` with `CitationScanner`. `citation_aside_html()` sits in `_citations.py`.
- Shared: `citation_scan` (21 cases) pins the projected text, `citation_holdback` pins what each `feed()` may emit. Every case is fed whole, at the boundaries the fixture names, and under 1000 characters at every split point and one character at a time, so chunk invariance is checked per case rather than as a test of its own.
- R: 22 scanner tests replaced by two runners over the new shared test definitions.

*Intentional gap at this point*: R resolves the aside's icon URLs from its served asset bundle, and Python has no bundle until `www/` becomes a root source (D10, milestone 8), so the Python aside omits the `icon` attribute and the title image. That is the branch R takes when the assets are absent. The fixture renders each aside per package through an `{{aside:N}}` token and pins every other character exactly. Once the Python UI stuff lands, this can be made fully consistent across both implemenations.

Testing: R 516 tests and 6639 expectations, no failures. Python 100 tests, `ruff` and `pyrefly` clean. The fixture expectations were derived from `R/citation-scan.R` by hand and passed against R on the first run.

## Quick R-side summary

No R source changed. `pkg-r/R/` is identical to the base branch, so there is no behaviour delta and no call site to check. The whole R change is `test-citation-scan.R`.

22 hand-written scanner tests are replaced by two runners over the new `citation_scan` and `citation_holdback` fixture sections. Each maps onto a named case, except the three that were about chunking rather than about a particular input: the runner now feeds every case whole, at the boundaries the fixture names, and under 1000 characters at every split point and one character at a time. The bounded-buffer test stays, because it asserts on the closure's `buf` rather than on output.

Evidence it is safe: I derived the expected strings from `R/citation-scan.R` and ran the R runner only after Python passed. It passed first try, so R's output matches an independent reading of it rather than a fixture fitted to it. Full suite 516 tests, 6639 expectations, no failures.

### DIFF
--- a/pkg-py/src/commons/_citation_scan.py
+++ b/pkg-py/src/commons/_citation_scan.py
@@ -75,6 +75,7 @@ class _Event:
 
 
 def _find(literal: str, buffer: str) -> re.Match[str] | None:
+    """Locate the first occurrence of a reserved literal in the buffer."""
     return _LITERALS[literal].search(buffer)
 
 
@@ -161,6 +162,7 @@ class CitationScanner:
         return self._step_discard()
 
     def _step_text(self) -> bool:
+        """Emit plain text up to the next reserved literal, or up to the holdback."""
         event = self._find_text_event()
         if event is not None:
             prefix = self._buffer[: event.pos]
@@ -185,6 +187,7 @@ class CitationScanner:
         return False
 
     def _step_citation(self) -> bool:
+        """Close, abandon, or keep buffering the citation body being collected."""
         event = self._find_reserved_event()
         if event is not None and event.action == "close" and event.pos <= self._cap:
             body = self._buffer[: event.pos]
@@ -204,6 +207,7 @@ class CitationScanner:
         return False
 
     def _step_discard(self) -> bool:
+        """Drop buffered text through the close tag of the element being discarded."""
         assert self._discard_close is not None
         match = _find(self._discard_close, self._buffer)
         if match is not None:
@@ -220,6 +224,7 @@ class CitationScanner:
         return False
 
     def _find_text_event(self) -> _Event | None:
+        """Find the earliest reserved literal that matters while in text mode."""
         pattern = (
             _CITATION_OPEN_AT_LINE_START
             if self._at_line_start
@@ -241,6 +246,7 @@ class CitationScanner:
         return min(events, key=lambda event: event.pos, default=None)
 
     def _find_reserved_event(self) -> _Event | None:
+        """Find the earliest reserved literal inside a citation body."""
         found: list[tuple[re.Match[str] | None, int, _Action]] = [
             (_find(CITATION_OPEN, self._buffer), len(CITATION_OPEN), "drop"),
             (_find(CITATION_CLOSE, self._buffer), len(CITATION_CLOSE), "close"),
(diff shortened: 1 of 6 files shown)


jat255:
## R-side summary

No R source changed. `pkg-r/R/` is byte-identical to the base branch, and `citation_scanner()`, `render_citation_aside()`, `citation_aside_html()`, and `project_citation_text()` all behave exactly as before. The entire R change is in one test file, so there is no behaviour delta to describe and no call site to check.

What moved, and why. The scanner's observable behaviour is something the two packages have to agree on, so it belongs in `tests/shared/` rather than being described twice in two languages. The scanner tests in `test-citation-scan.R` were prose descriptions of that behaviour written in R. They are now two runners over `citation_scan` and `citation_holdback` in the shared fixture, which the Python suite reads as well.

The 22 tests that went, and what replaced them. Every one of them maps onto a named case in `citation_scan`: the pass-through and code-block and table cases, the verified and unverified and malformed citations, the case-insensitive tags, the line-start anchor on the opening tag, the unterminated markup, the three cap cases, the recovery cases after oversized and malformed elements, and the two nesting cases. The three that were about chunking rather than about a particular input (`chunk-invariance: any character split yields identical output`, `opening and closing tags work in one-character chunks`, and `adversarial multi-element input is invariant to random partitions`) have no case of their own, because the runner now feeds every case whole, at the boundaries the fixture names, and for texts under 1000 characters at every split point and one character at a time.

The one test that stayed is `an oversized unclosed citation keeps only a bounded close-tag suffix`. It reaches into the closure with `get("buf", envir = environment(scanner$feed))`, so it is about how much the scanner retains rather than what it emits. That is not a cross-language contract and it should not become one.

Why this is safe, and what was checked. The fixture's expected strings were not generated from R. I derived each one by reading `R/citation-scan.R`, wrote them into the fixture, implemented the Python scanner against them, and only then ran the R runner. It passed on the first run, so R's actual output matches what I had written down independently. Had I generated the expectations from R, a passing R suite would have proved nothing. The full R suite is 516 tests and 6639 expectations with no failures.

Where the coverage is genuinely stronger. Several of the old tests asserted with `expect_match()` on substrings, which left the rest of the projected string unchecked. `scanning resumes after oversized and malformed citations` and `nested model markup is removed without leaking later markup` are the clearest examples: they checked that certain fragments appeared and that `<commons-citation` did not, but not the exact separators between them. The fixture pins the entire output string for each case.

Where it is different rather than stronger. Randomness is gone. The old `adversarial multi-element input` test cut its fixture at 50 random points per run under a fixed seed. The new coverage is exhaustive on short cases, which is strictly more, but the three cases whose bodies reach the 16 KB element cap are too long to split exhaustively, so they name their own boundaries instead. Those boundaries straddle the opening tag, the closing tag, and the point where the body crosses the cap. That is the part worth a second opinion: if you think a long case wants random partitions on top of the named boundaries, say so and I will add them.

Also worth your scrutiny. The fixture runner in `test-citation-scan.R` carries more machinery than a test file usually should, specifically `citation_scan_padding()`, `citation_scan_expected()`, and `citation_scan_chunkings()`. The padding helper exists so a 16 KB case costs one integer on disk rather than 16 KB of `x`. The expected helper exists because the R aside carries icon URLs the Python one cannot resolve yet, so each package renders its own aside for the `{{aside:N}}` token. Both are the price of one fixture instead of two test suites, but I would rather hear now if the indirection reads as too much.

jat255:
@simonpcouch this one could also use a quick look, but has minimal (only test suite) changes on the R side.

simonpcouch (approved):
Nice!

PR #240, opened 2026-09-02
Python context layer: construction and frontmatter handling

jat255:
First of two PRs for M4, the Python context layer (kata [`63jx`](https://www.katatracker.com/)). This one adds construction and text handling; [#241](https://github.com/posit-dev/commons/pull/241) adds retrieval on top.

`context_layer(files=...)` reads its files eagerly and strips YAML frontmatter, so a bad path fails at construction rather than mid-conversation, and maintainer metadata such as a provenance URL never reaches the index. The returned `ContextLayer` is a plain class per D8. `commons.__all__` is no longer empty.

The frontmatter regex is anchored to the start of the document and consumes only the first fence, so a `---` thematic break in the body survives. The metadata block is optional, so a fence whose keys have been deleted is removed rather than indexed as literal `---` text.

`tests/shared/context_layer.json` pins the nine stripping cases, and both suites run them. Retrieval ranking is deliberately not in the fixture: the two BM25 engines score differently, and per the D7 refinement that is implementation detail rather than observable behaviour.

## R changes

Only one functional change in `pkg-r/R/context-layer.R` to correct a small (potential) bug. Previously, [`strip_frontmatter()`](https://github.com/posit-dev/commons/pull/240/changes#diff-e528236cc355dfc99f15a7aaed050dfdabc0121d09b8420942e84215bfe957dcR107) would not remove a document opening with `---\n---` (i.e. it had a frontmatter section, but it was empty).

The regex was `^---\r?\n.*?\r?\n---(\r?\n|$)` and is now `^---\r?\n(.*?\r?\n)?---(\r?\n|$)`. Making the metadata block optional is the whole change. Previously a document opening with `---\n---` did not match, so the fence was left in place; `context_layer()` then saw non-empty text, kept the document, and indexed `---\n---` as a chunk. Now it matches and the fence is removed, which for a fence-only file means the document is dropped entirely.

Blast radius is small. `strip_frontmatter()` has one caller, `context_layer()`, and its only other reader is the new test. The change cannot widen what gets stripped, because adding an optional group to a lazy pattern only lets it match where it previously failed, and the anchor is unchanged.

Evidence: the full R suite is green at 6649 passing, 0 failed, 0 errors, 28 skipped. The nine shared cases were run against the R implementation directly and against the Python one, and both satisfy all nine.

No hand-written R tests were deleted. The new test is additive. `pkg-r/tests/testthat/fixtures/shared/context_layer.json` is generated by `scripts/sync-shared-fixtures.sh` and needs no review.


pkg-r/tests/testthat/test-context-layer.R
```diff
@@ -1,3 +1,17 @@
+test_that("strip_frontmatter matches the shared cases", {
+  cases <- shared_fixture("context_layer")$strip_frontmatter$cases
+  # An empty list would make the loop below vacuously succeed.
+  expect_gt(length(cases), 0)
```
simonpcouch:
```suggestion
```

Not really worried about this

simonpcouch (approved):
🏄

pkg-r/tests/testthat/test-context-layer.R
jat255: Even though it's vacuous? ;)

simonpcouch: lmaooooo

PR #255, opened 2026-09-02
One conformance corpus and one definitions contract for both packages

jat255:
## Summary

Fourth PR of M2 (data layer). Gives the two packages a single shared copy of the data-dict definition fixtures — 14 YAML files of valid and invalid table definitions written in data-dict's expression language, which both test suites use to check their exporters against the reference `data-dict` binary — and a shared fixture (`tests/shared/definitions.json`) pinning the expected export for every definition, so both packages are held to the same contract. Also includes some groundwork for the definitions registry and the compiler, which land next.

## The definition fixtures move out of `pkg-r`

The 14 YAML fixtures were moved out of `pkg-r/tests/testthat/fixtures/` to now live in `tests/shared/definition-export/`. Both suites read the same files. 

## `definitions.json`

This file pins behavior on three things, in three different sections within the file:

- `export_records` — the expected export for each of the 42 valid definitions: its SQL translation, its inferred kind and type, and the columns and definitions it references. Generated from the real `data-dict` binary at pinned commit `d950c5a`; the binary exports more than commons needs, so the generator keeps only the fields both packages consume (the "projection"). Regenerate with `scripts/generate-definitions-fixture.sh` rather than editing by hand.
- `mixed_grain` — a per-definition boolean: true when a definition's exported shape is `row` but its expression tree contains an aggregate, directly or through a definition it references, which is the combination `call_metrics`' mixed-grain guard rejects. The flag is absent from data-dict's export — it exists only in the typed IR, data-dict's internal type-annotated parse tree — so it cannot be generated from the binary. It is hand-maintained (computed with the R implementation's grain logic) and the generator preserves it. One of the 42 definitions is mixed-grain, and a test asserts the values are not all identical: an all-false fixture would pass against an implementation that always answers false, pinning nothing.
- `invalid` — maps each of the 11 invalid fixtures to the specific data-dict problem code a conforming compiler must report for it (`cycle.yaml` must fail with the cycle error, not a generic parse failure), and both suites assert on that code. Replaces a literal map hard-coded in the R helper, which Python could not share.

## The generator refuses to run against the wrong binary

`scripts/generate-definitions-fixture.sh` resolves the cargo-installed binary by path rather than trusting `PATH`, and verifies that installation is the pinned revision. A fixture generated from another build would bless whatever that build does, in a way nothing downstream could detect.

## Only R asserts the contract so far

There are no tests on the Python side for the actual data-dict definition contract in this PR (since that implementation does not exist yet). The only python tests are for the mechanics of loading the definition fixtures.

## Verification

219 R assertions, none skipped. The new R assertions were then checked by perturbing one translation and one grain value in the fixture and confirming the suite fails.

---

## Summary of R changes

No package code changed. Everything here is under `pkg-r/tests/`, so no exported or internal behaviour moves. Files under `pkg-r/tests/testthat/fixtures/shared/` are generated by `scripts/sync-shared-fixtures.sh`.

**What changed and why.** The definition-export fixtures used to live inside the R package's test fixtures, where only R could reach them. Python needs the same files to check its compiler against the same cases, and a second copy would drift, so the fixtures moved to the repository root and R now reads the synced copy under `fixtures/shared/`, exactly as it already does for the provenance and citation fixtures. Alongside that, the map from an invalid fixture to its expected data-dict problem code moved out of the helper and into the shared fixture, because it is a fact about the shared cases rather than about R.


jat255:
@simonpcouch this one's ready for a quick look on the R side. No changes to actual running code. Like the last few PRs, it's a restructuring of the test cases into shared fixtures that can be used on the python side. Thanks!

pkg-r/tests/testthat/helper-data-dict.R
```diff
@@ -31,30 +31,26 @@ data_dict_cli_context <- function(result) {
   )
 }
 
+# The conformance corpus is a cross-language fixture: both suites read the
+# same YAML files. These read the synced copy, as every shared fixture does.
```
simonpcouch:
```suggestion
```

Seems well-documented at this point

pkg-r/tests/testthat/helper-data-dict.R
```diff
@@ -31,30 +31,26 @@ data_dict_cli_context <- function(result) {
...
     "fixtures",
+    "shared",
     "definition-export",
     kind,
     "*.yaml"
   )))
 }
 
+# Which data-dict problem code each invalid fixture must produce. Shared, so
+# both suites hold one copy of the mapping.
```
simonpcouch:
```suggestion
```

pkg-r/tests/testthat/helper-data-dict.R
```diff
@@ -90,3 +86,45 @@ definition_export_contract <- function(export) {
   }
   out
 }
+
+# The fixture arrives from JSON as nested lists, while
+# definition_export_contract() produces character vectors. Normalize the
+# fixture side so a comparison reports a real difference rather than a
+# difference in how each format spells an empty sequence.
```
simonpcouch:
```suggestion
# definition_export_contract() produces character vectors.
```

pkg-r/tests/testthat/test-definition-export.R
```diff
@@ -179,3 +179,39 @@ test_that("invalid fixtures also fail an installed data-dict", {
...
+test_that("the export matches the shared definitions contract", {
+  skip_if_not_installed("yaml")
+  spec <- shared_fixture("definitions")
+  paths <- definition_fixture_paths("valid")
+  expect_gt(length(paths), 0)
+
+  for (path in paths) {
+    export <- definition_export_spec(yaml::read_yaml(path))
+    # Keyed contract, not a sequence: the generated file sorts its keys for
+    # stable diffs while the export keeps authored order.
```
simonpcouch:
```suggestion
```

simonpcouch (approved):
Nice!

pkg-r/tests/testthat/test-definition-export.R
jat255: Thanks for these comments. Even with my pruning, claude is so overly verbose these days...

PR #257, opened 2026-09-02
Show plots to users in a higher resolution

### LINKED ISSUE #238: Graphs appear low-resolution in the UI
Maybe the user sees a higher-resolution version than the model? Though don't want to introduce a situation where the model sees some plotting artifact that the user doesn't—we should start with the high-resolution and then downscale rather than rendering twice.

simonpcouch:
Closes #238.


simonpcouch:
Before:

Screenshot 2026-09-10 at 8 05 19 AM


After:

Screenshot 2026-09-10 at 8 04 30 AM

PR #268, opened 2026-09-04
Cancel run_r's later callbacks so the loop drains

### LINKED ISSUE #267: run_r leaks later timers, which stalls the R test suite for ten minutes
`R CMD check` on `pkg-r/` spends about ten minutes idle. [Run 33796849087](https://github.com/posit-dev/commons/actions/runs/33796849087/job/100786518215) took 15m22s, and the `checking tests` step was 11m25s of that. Check reports the split itself:

```
Running 'testthat.R' [2m/11m]
```

Two minutes of CPU, eleven minutes of wall clock. Every other check step takes a few seconds. The same stall happens locally, where the suite looks hung on `test-trajectory-review.R`.

## Cause

`shiny::testServer()` reads an output through `MockShinySession$getOutput()`, which calls `shiny:::extract()` and then `shiny:::wait_for_it()`:

```r
wait_for_it <- function() {
  while (!later::loop_empty()) later::run_now(0.1)
}
```

That does not wait for the output's own promise. It waits until the whole global `later` queue is empty, so anything left on the queue has to fire first.

`run_r_tool()` leaves things on that queue. Every call ends in [`schedule_worker_reap()`](https://github.com/posit-dev/commons/blob/76f9e38edcf42a19e86245a05f8740ee342e5793/pkg-r/R/run-r.R#L433), which arms a timer at `commons.run_r_idle_timeout + 1`, 601 seconds by default. Each call also arms a timeout timer in [`worker_await()`](https://github.com/posit-dev/commons/blob/76f9e38edcf42a19e86245a05f8740ee342e5793/pkg-r/R/run-r.R#L471) at `commons.run_r_timeout`, 60 seconds by default. Neither is cancelled when the call settles, and `worker_close()` does not cancel them either.

So `test-run-r.R` finishes in ten seconds and hands the global loop 70 pending callbacks. `test-trajectory-review.R` then reads `output$entries` inside a `testServer()` block ([test-trajectory-review.R#L297](https://github.com/posit-dev/commons/blob/76f9e38edcf42a19e86245a05f8740ee342e5793/pkg-r/tests/testthat/test-trajectory-review.R#L297)), and that read cannot return until the last 601 second timer fires.

## Measurements

Immediately after `test-run-r.R`, the global loop holds 70 callbacks in two clusters: 34 at 46 to 60 seconds, and 35 at 587 to 601 seconds. The two clusters match the two option defaults. The farthest is 600.9 seconds out, and draining the loop the way `wait_for_it()` does takes 600.9 seconds.

Running `test-run-r.R` and then `test-trajectory-review.R` in one session, the test "the viewer filters conversations and follows selection" takes 601.1 seconds. With `commons.run_r_idle_timeout = 2` and `commons.run_r_timeout = 2`, the same test takes 3.1 seconds. A bisect over every other test file shows `test-run-r.R` is the only one that triggers this. Run on its own, the whole suite is about 40 seconds.

A stack sample of the stalled process shows it parked in `pthread_cond_timedwait` at about 0% CPU, which is why the job burns wall clock and not CPU.

## Fix

`later::later()` and `later::later_fd()` each return a canceller function. Keep those handles on the worker, call them when the call settles, and call them again in `worker_close()`. Note that `later::cancel()` is not exported and does not exist in later 1.4.8.

The comment above `schedule_worker_reap()` says stray timers are harmless because they check recency before acting. That holds for correctness, but not for any caller that waits on `later::loop_empty()`.

Only `MockShinySession$getOutput()` reaches `wait_for_it()`, so deployed apps do not stall. The leak itself is still real there.

jat255:
Fixes #267.

> To be completely honest, this is touching R code and mechanics in a way I totally do not understand, so this fix is totally on Claude. @simonpcouch please take a look over the approach and double-check it makes sense? Thanks!

Every `run_r_tool()` call armed two `later` callbacks and cancelled neither, and `worker_close()` didn't either. `shiny::testServer()` spins until `later::loop_empty()`, so after `test-run-r.R` the loop held 70 callbacks, the farthest 601 seconds out, and the first `testServer()` block in `test-trajectory-review.R` blocked for 601.1s. That was most of the 15-minute CI job.

`later()` and `later_fd()` return canceller functions (there is no `later::cancel()`). The change, all in `pkg-r/R/run-r.R`:

- Keep cancellers for the call timeout, kill escalation, and `later_fd()` watcher in `worker_await()`, and release them all from `settle()`: a finished call leaves nothing on the loop.
- Hold one reap timer per worker instead of one per call, and cancel it in `worker_close()`. Reap timing and target selection are unchanged.
- Interrupt and kill paths are untouched: callbacks are cancelled only once `settled` is TRUE, at which point they would have returned early anyway.

Verification: the new test in `test-run-r.R` fails before the change and passes after; the viewer test went from 601.1s to 0.65s; `R CMD check` is 1m17s, `Status: OK`, no notes. Full suite passes locally in 52s (28 skips, all environmental). No tests deleted; one added.

The CI tests drop from about 15m for the "R CMD check" job to ~5m.

Worth your scrutiny:

- The ordering in `poll()`, where the callback clears `cancel_watch` and may re-arm it through a recursive `poll()` call. I am confident it is right but it is the fiddliest part.
- Whether cancelling the reap timer on every call, rather than letting redundant timers pile up and no-op, is the behaviour you want for a long-lived worker under steady load.

### DIFF
--- a/pkg-r/R/run-r.R
+++ b/pkg-r/R/run-r.R
@@ -279,6 +279,7 @@ new_r_worker <- function(network = "none", protection = "sandbox") {
   worker$synced <- 0L
   worker$tail <- NULL
   worker$pending <- 0L
+  worker$reap <- NULL
   worker$last_used <- Sys.time()
   worker$runtime <- worker_runtime()
   # Close the child process when the agent is garbage-collected, so a session
@@ -369,6 +370,7 @@ commons_dll_path <- function() {
 }
 
 worker_close <- function(worker) {
+  cancel_worker_reap(worker)
   if (!is.null(worker$rs)) {
     try(worker$rs$close(), silent = TRUE)
     worker$rs <- NULL
@@ -424,14 +426,18 @@ worker_call <- function(runtime, name, args) {
 }
 
 # Close the worker after a quiet stretch; it respawns lazily on the next
-# call. Stray timers are harmless: they check recency before acting.
+# call. One timer per worker, not one per call: shiny::testServer() reads its
+# outputs through shiny:::wait_for_it(), which spins until the whole later
+# queue drains, so a stray timer stalls the reader for its full delay.
 schedule_worker_reap <- function(
   worker,
   idle = getOption("commons.run_r_idle_timeout", 600)
 ) {
   worker$last_used <- Sys.time()
-  later::later(
+  cancel_worker_reap(worker)
+  worker$reap <- later::later(
     function() {
+      worker$reap <- NULL
       quiet <- difftime(Sys.time(), worker$last_used, units = "secs") >= idle
       if (worker$pending == 0L && quiet) {
         worker_close(worker)
@@ -439,6 +445,15 @@ schedule_worker_reap <- function(
     },
     delay = idle + 1
   )
+  invisible(worker)
+}
+
+cancel_worker_reap <- function(worker) {
+  if (!is.null(worker$reap)) {
+    worker$reap()
+    worker$reap <- NULL
+  }
+  invisible(worker)
 }
 
 # Resolve when the in-flight $call() completes, without blocking: later_fd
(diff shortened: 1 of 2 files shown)


simonpcouch (commented):
I would have thought that this could be achieved without changes in R/, and the double assignment gives me the heebie jeebies a bit. Sorry to not approve and also not have the bandwidth to really look into this—planning to spend some time on commons tomorrow and will see what I can do here then!

simonpcouch:
In the meantime, feel free to ignore the R CMD check GHA.

jat255:
> I would have thought that this could be achieved without changes in R/, and the double assignment gives me the heebie jeebies a bit. Sorry to not approve and also not have the bandwidth to really look into this—planning to spend some time on commons tomorrow and will see what I can do here then!

Totally understandable. I'll ask Claude to take another look to see if it's fixable on the test side only. It seemed convinced this was the right way, but I also didn't push back.

simonpcouch (approved):
F****** ASYNC

Okay, yes, I'm on board. Thanks for your patience :) Just sent up some small changes to avoid some of the double assignments and simplify cancellation.

Thanks for catching this! I see that tests now wrap up in ~50s or so locally. We could also test in parallel--will file an issue to do so.

PR #270, opened 2026-09-04
Type-check definition expressions and resolve their references

jat255:
Second of four PRs porting the data-dict definition compiler to Python (kata f6hz, stage 1). Stacked on #269. Turns a parsed expression into an export record: inferred type and kind, direct column and sibling-definition references, and the checks that refuse a bad definition at construction.

`translations` stays empty by design. The DuckDB emitter is the next PR, so this PR asserts `kind`, `type`, `columns` and `definitions` for all 42 corpus definitions and leaves the SQL to #3.

Verified against the installed `data-dict` binary, not only against the fixture, with the test skipped when the binary is absent. The corpus tests were confirmed to bite by perturbing kind inference, reference deduplication, and the column-shadowing check.

## Regular expressions took six review rounds, and the design changed twice

data-dict validates patterns with Rust's `regex`. My first version validated with Python's `re`, which is not a proxy for it in either direction: `re` accepts lookaround and backreferences Rust rejects, and rejects `\p{L}`, `\z` and POSIX classes Rust accepts. Each round of review found another divergence, and each targeted fix produced the next one.

So it validates and matches with RE2, which shares Rust's finite-automata design and agrees with it on every case raised across those reviews. The first cut reached RE2 through the `duckdb` already in commons' dependencies; review then surfaced `google-re2`, Google's own binding, maintained alongside RE2 itself and shipping wheels for the whole supported matrix. That deleted the shared connection, its lock, and the deferred import, along with the read-time coupling to DuckDB that an earlier version of this section flagged for reviewer judgement.

Two constructs are known to differ, both verified against the binary as accepted by data-dict and refused here: extended mode `(?x)` and CRLF-aware multiline `(?R)`. Rust's named-group spelling `(?<name>...)` was a third, but google-re2's RE2 is newer than DuckDB's vendored copy and has accepted that spelling since 2023-07, so the divergence and its error-message hint are gone. The tests deliberately do not claim the list is exhaustive; an earlier version did and was wrong. Both remaining differences fail closed at construction with a message, and neither changes what a definition matches. Closing the gap needs a binding to Rust's `regex`, and the only Python one is unmaintained.

## Departures from R

Integer literals are `int` rather than the normalized string R keeps for 64-bit precision, with the i64 range check retained. Where R reaches PCRE for regexes, this uses RE2, so the two accept slightly different pattern sets at the edges described above.

Verification: 483 tests pass, ruff and pyrefly clean.


jat255:
Approved after swapping to `google-re` at 11PM

PR #271, opened 2026-09-04
Lower type-checked definitions to DuckDB SQL

jat255:
Third of four PRs porting the data-dict definition compiler (kata f6hz, stage 1). Stacked on #270.

This is where conformance completes. `translation.code` is now asserted byte for byte and `notes` exactly, for all 42 corpus definitions, against the shared fixture and against an installed `data-dict` binary.

Numbers are formatted by widening the significant digits until the text reads back as the same double, which is the R algorithm rather than `repr()`. `repr()` switches to exponent notation and would emit `1e+20` where data-dict emits the digits. Perturbing the formatter to `repr()` fails only the dedicated boundary test and leaves the 42-definition corpus green, so that test earns its place: the corpus never reaches the case.

LIKE lowers to the narrowest form an authored pattern allows, so a wildcard-free pattern becomes an equality and a single anchored wildcard becomes a prefix or suffix test. A pattern that is itself an expression stays a LIKE, because none of that is knowable at compile time.

The notes are why the emitter is hand-written rather than generated through sqlglot: they are collected as the writer walks the IR, and sqlglot has no way to produce them.

`_export.py` imports the emitter inside `_resolve_one` rather than at module scope, because the emitter needs `Ir` from `_export`. Worth a look if you would rather the IR moved to its own module to break the cycle properly.

Verification: 545 tests pass, ruff and pyrefly clean.

### DIFF
--- a/pkg-py/src/commons/_definitions/_emit_duckdb.py
+++ b/pkg-py/src/commons/_definitions/_emit_duckdb.py
@@ -1,0 +1,345 @@
+"""Lowering a type-checked expression to DuckDB SQL.
+
+A port of data-dict's DuckDB emitter. Only DuckDB is ported, because commons
+does not consume data-dict's R targets, and conformance against the data-dict
+binary is the authority for what this must produce.
+
+The emitter also collects notes: places where DuckDB's semantics differ from
+data-dict's, which the model is told about rather than silently inheriting.
+"""
+
+from __future__ import annotations
+
+import math
+import re
+from typing import Any
+
+from ._export import Ir
+
+__all__ = ["emit_duckdb"]
+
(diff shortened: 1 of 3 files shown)


jat255:
Reviewed 2026-09-04 23:05 MDT — manual review complete. Findings: 3 MEDIUM (circular-import workaround avoidable via the codebase's TYPE_CHECKING idiom; stale `translations` docstring/comment; small-magnitude float notation unpinned by tests), 1 LOW (emitter errors bypass `translation["error"]`), no blockers. All three MEDIUMs fixed in 9daded3; full suite green (545 passed), roborev clean.

PR #272, opened 2026-09-04
Bind definitions to a source and wire both compiler phases

jat255:
Last of four PRs porting the data-dict definition compiler (kata f6hz, stage 1). Stacked on #271. With this, a governed definition authored in data-dict's expression language reaches the agent as executable SQL.

Phase 1 runs when a dictionary is read, so an unusable definition fails at `data_dictionary()` before any source exists. Phase 2 runs at `data_source()`, where a dialect is finally known, and fills the `compiled_definitions` the registry has been rendering since #261. That is the same split R uses, so both packages report a bad definition at the same constructor.

Composition inlines a definition's sibling references and merges their notes. Each reference is renamed to a unique marker before the SQL is emitted, rather than searching the emitted SQL for the definition's own name afterwards, which is what makes a collision impossible. The substitution scans rather than string-replaces, so a string literal holding a marker is left alone.

Grain comes from the typed IR because the exported kind cannot answer it: a row expression can hold an aggregate child without becoming an aggregate. A metric whose chain mixes grain is refused, since one SQL expression cannot express the subquery rewrite it would need.

Only DuckDB is lowered, so a source of any other dialect is refused at construction rather than emitted for wrongly. Snowflake and Databricks are stage 2 of the issue.

## Verification

End to end against a real DuckDB source: the compiled SQL is executed, not just compared with a fixture, and expanded `{{token}}` queries return the right rows. Every corpus definition is also parsed by DuckDB.

That parse check is worth a reviewer's eye. `json_serialize_sql` reports a parse failure in its returned payload rather than raising, so the first version of the test passed unconditionally. It now asserts the payload, and was confirmed to fail when the emitter leaves a `CASE` unterminated.

`export_spec()` now accepts dictionary entries as models as well as raw mappings, because callers construct `DataDictionary` both ways and phase 1 runs for both. That was a regression caught by the existing registry tests.

568 tests pass, ruff and pyrefly clean.

## R changes

Nothing under `pkg-r/R/` changes here. R runtime behavior is identical before and after, and the only R changes are in the test suite to use shared fixtures.


jat255:
Reviewed 2026-09-05 21:20 MDT (final manual review pass).

Findings from the review were addressed in 966cb21 and 0e56cab:

- Definition compilation is now atomic (a refusal leaves the dictionary untouched) and the exposed-table set is required, so the unexposed-table check can't be skipped by omission.
- Tests now pin the marker machinery itself: a string literal holding marker text and a column named like a marker are both left alone.
- **The shared-fixture work caught a real bug:** references inside CASE branches were never inlined (the IR walkers didn't see mappings inside lists), so e.g. the corpus's own `enterprise_revenue` compiled to SQL referencing a column that doesn't exist. Fixed, and `tests/shared/definitions.json` now has a `composed` section pinning composed SQL + merged notes for all 42 corpus definitions, executed by both suites (R and Python agree on every case).
- Docstrings now record where reading a dictionary / constructing a source can fail.

Checks on the pushed head: ruff clean, pyrefly 0 errors, 689 tests pass; the new R fixture runner passes (85 expectations).

simonpcouch (approved):

PR #273, opened 2026-09-04
Lower governed definitions to Snowflake and Databricks

jat255:
Stage 2 of the definition compiler (kata yfan), split out of f6hz once stage 1 landed. Stacked on #272. Ports `pkg-r/R/definition-emit-sql.R`, so a source on either warehouse compiles its definitions instead of being refused at construction.

## These targets have no upstream authority

Every other part of the compiler is checked against the pinned `data-dict` binary. These emitters cannot be: the binary emits `SQL(duckdb)` and three R targets, and nothing else. So there is no output to generate from and nothing external that says what correct looks like. kata g0ax is the upstream ask that would let one Rust implementation serve both languages and delete both emitters.

The expectations instead live in `tests/shared/definition-warehouse-sql.json`, hand-maintained, keyed to the same corpus dictionaries as the other definition fixtures, and read by both suites. Cases pin `code` exactly where the whole string is worth reading, and `code_contains` or `notes_contain` where a rule is the point and the full rendering would bury it.

Both runners were confirmed to bite by perturbing one expected string in the fixture and watching each suite fail.

## The port

Snowflake and Databricks share precedence and most syntax, so they are one writer parameterised by dialect, matching the R structure. The differences are concentrated where lowering has to be target-aware: identifier quoting, struct access, datetime construction, guarded division and modulus, non-finite literals, and regular expressions.

A construct a dialect cannot express raises inside the walk and is returned as an error on the translation rather than propagating, so one definition that cannot be lowered does not stop the others. Only the chosen target's error fails construction.

## Verification

578 Python tests, ruff and pyrefly clean. Full R suite 6588 passing, 0 failures.

## R changes

The R package gains no new behaviour here. `R/` is untouched. The change is to `tests/testthat/test-definition-compile.R`, and it is a test-provenance change rather than a behavioral one.


jat255:
Reviewed 2026-09-05 21:24 MDT — manual review complete. Findings: 1 HIGH (Python had no coverage of the unsupported-construct → construction-failure path), 3 MEDIUM (silent single-target vs all-targets parity gap; shared fixture never exercised composition/marker substitution; mistyped fixture keys would pass while asserting nothing), 1 LOW (kata reference dressed as a GitHub issue link in tests/shared/README.md). All fixed in 73d7563 + c457148: error cases and a composition case now live in the shared fixture and run in both suites (the two hand-written R test blocks they duplicated are removed), both runners reject unknown expectation keys, the parity difference is documented in _compile.py, and the README no longer references the dev-local ledger. The parity gap itself is accepted as an implementation difference. Python: ruff/pyrefly clean, 696 passed; fixture verified to bite by perturbation. R runner changes are parse-checked but could not run locally (local R library has ellmer 0.4.2 < required 0.5.0) — CI is the verification there. roborev clean on both commits.

simonpcouch (approved):

PR #274, opened 2026-09-04
Give TableId a catalog component

### LINKED ISSUE #302: R reads a three-part table name string as a dotted schema instead of catalog.schema.table
## Bug

In R, `data_source(con, tables = "ANALYTICS.PUBLIC.ORDERS")` folds everything before the last dot into the schema: `DBI::Id(schema = "ANALYTICS.PUBLIC", table = "ORDERS")`. Quoted, that is `"ANALYTICS.PUBLIC"."ORDERS"` — one identifier with a dot inside it, naming a schema no warehouse has. The zero-row probe then fails and `data_source()` errors with a "table does not exist" for a table that does.

```r
id <- DBI::Id(schema = "ANALYTICS.PUBLIC", table = "ORDERS")
DBI::dbQuoteIdentifier(DBI::ANSI(), id)
#> <SQL> "ANALYTICS.PUBLIC"."ORDERS"
```

The fold dates to 97db857 ("support schema-qualified connection tables", closes #15), written before the package had any catalog support. Catalogs arrived later via `DBI::Id(catalog = ...)` objects (the Snowflake/Databricks work), and the string parser was never revisited.

R also accepts `"a.b.c.d"`, folding it into a doubly-dotted schema.

## Expected

Match the behavior #274 gives the Python side:

- `"catalog.schema.table"` → `DBI::Id(catalog = ..., schema = ..., table = ...)`
- More than three parts → an error naming the input and pointing at `DBI::Id` for literal names containing dots
- One and two parts behave as today

The workaround meanwhile is an explicit `DBI::Id(catalog = "ANALYTICS", schema = "PUBLIC", table = "ORDERS")`.

## Shared contract

Table-name parsing is user-observable behavior both packages expose through `tables`, so the parsing rules belong in `tests/shared/` as an executable fixture read by both suites. The fixture cannot land until R is fixed, because R currently fails the three-part case. Being fixed in #274.

jat255:
First PR of the catalog import stack (kata xp65). Stacked on #273.

## Summary

A warehouse names a table `catalog.schema.table`, and `TableId` had two levels, so everything before the last dot became the schema. `ANALYTICS.PUBLIC.ORDERS` quoted to `"ANALYTICS.PUBLIC"."ORDERS"`: one identifier with a dot inside it, naming a schema no warehouse has. That is a live bug for engine sources today, not only a gap ahead of catalog import, which is why it is a fix rather than a feature.

Three levels now, kept apart rather than folded into a string. `parts` is what quoting and probing walk, so neither has to know how many levels a backend has.

Two inputs are now refused rather than accepted. A `TableId` with a catalog but no schema, because there is no level between them to leave out. And a name with more than three parts, which previously became a schema containing dots.

The engine inspector joins the levels above the table into one dotted `schema`, which is SQLAlchemy's convention, rather than passing the schema alone and probing the wrong namespace. No dialect available to the suite has three levels, so that argument is asserted by recording it rather than by a real probe, which is the one piece of test-double here and the comment says why.

---

## R Changes

@simonpcouch there are some R behavior changes in this PR, to address an issue with how catalogs are handled on the R side, filed as #302 and fixed here. R's string parser now reads three parts as `catalog.schema.table` and refuses more than three, matching the Python behavior above. The shared fixture caught R's `strsplit()` silently dropping a trailing empty piece, so `"orders."` parsed as a bare name; a trailing dot is now checked separately.

The parsing rules are user-observable in both languages, so they live in `tests/shared/table-names.json` with a runner in each suite, replacing the per-language copies added earlier in this stack.

### DIFF
--- a/pkg-py/src/commons/_backends.py
+++ b/pkg-py/src/commons/_backends.py
@@ -65,8 +65,7 @@ class DuckDBBackend:
         return [row[0] for row in rows]
 
     def quote(self, table_id: TableId) -> str:
-        parts = [table_id.schema, table_id.table] if table_id.schema else [table_id.table]
-        return ".".join(quote_identifier(part) for part in parts)
+        return ".".join(quote_identifier(part) for part in table_id.parts)
 
     def dialect(self) -> str:
         return "duckdb"
@@ -96,9 +95,16 @@ class EngineBackend:
     def quote(self, table_id: TableId) -> str:
         preparer = self._engine.dialect.identifier_preparer
         quoted = preparer.quote(table_id.table)
-        if table_id.schema:
-            return f"{preparer.quote_schema(table_id.schema)}.{quoted}"
-        return quoted
+        if table_id.schema is None:
+            return quoted
+        # quote_schema() is given one component at a time: handed
+        # "ANALYTICS.PUBLIC" it produces one identifier containing a dot.
+        outer = ".".join(
+            preparer.quote_schema(part)
+            for part in (table_id.catalog, table_id.schema)
+            if part is not None
+        )
+        return f"{outer}.{quoted}"
 
     def dialect(self) -> str:
         return self._engine.dialect.name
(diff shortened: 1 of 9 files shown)


jat255:
Manual review completed 2026-09-05 23:04 MDT by @jat255. (Reviewed as part of the M2 stack pass; the R-side parsing bug found during review was filed as #302 and fixed here in 4e3f039.)

simonpcouch (approved):

PR #275, opened 2026-09-04
Merge a warehouse catalog into an authored dictionary

jat255:
Second PR of the catalog import stack (kata task `xp65`). Stacked on #274. Ports `catalog.R`: selection into labelled relations, exclusion globs and the object cap, the manifest that decides whether a listing belongs in the prompt or behind search, catalog search, and the dictionary merge.

The python side follows the same rules used on the R side: authored prose wins, warehouse types and nullability are authoritative, identifier case normalizes per backend, and an ambiguous relative name is an error rather than a guess. Discovered columns come first in the order the warehouse reported them; authored columns with no counterpart follow.

## R changes

@simonpcouch 

The changes on the R side are tests only — `pkg-r/R/` is untouched.

- `tests/shared/catalog-merge.json` is a new cross-language contract: the three catalog limits (object cap, prompt threshold, search probe bound), exclusion-glob semantics, and seven merge scenarios with expected merged tables, columns, and definition bindings.
- `pkg-r/tests/testthat/test-catalog-merge.R` is the R runner for it, covering the limits, `catalog_excluded`, and `catalog_merge_dictionary` against fixture-built relations. `pkg-r/tests/testthat/fixtures/shared/catalog-merge.json` is the synced copy from `scripts/sync-shared-fixtures.sh`.
- One behavior the fixture newly pins on both sides: exclusion globs match the bare table name, case-sensitively.
- The fixture's `kind` expectations match what `catalog_merge_dictionary_table` already does; adding `kind` to the merged table was a Python-side gap this PR closes.


jat255:
Reviewed 2026-09-05 23:24 MDT (final manual review pass).

Findings from the review were addressed in c494065:

- **Silent divergence fixed:** the merge dropped `Relation.kind` where R carries it onto the merged table; `Table` gains a `kind` field and the merge sets it (warehouse wins, authored survives as fallback).
- **Coverage gaps closed:** the `search()` `queryable` probe path and `merge_dictionary`'s `access_check` are now exercised, and the object-cap and prompt-threshold boundaries are pinned at exactly-at-the-limit.
- **Shared fixture added:** `tests/shared/catalog-merge.json` pins the three limits, exclusion-glob semantics (bare table name, case-sensitive — now documented in `excluded()`), and seven merge scenarios with expected tables, columns, and definition bindings. Both suites run it; the R runner passes all cases, and the three error cases were confirmed to raise the intended errors on the R side.
- Smaller parity fixes: duplicate-label refusal now reports every duplicated label; a non-whole `limit` is refused cleanly; raise conditions documented on the three public functions.

Deliberately not changed: exclusion globs still match the bare table name case-sensitively (R agrees, and the fixture now pins it — a schema-qualified or case-folding exclude would be a cross-package decision, flagged for follow-up). `authored_name` remains settable from authored YAML; stripping it in the parse path would break round-tripping a merged dictionary through model_dump/model_validate.

Checks on the pushed head: ruff clean, pyrefly 0 errors, 761 tests pass; the new R fixture runner passes (57 expectations). Note: the full R suite could not run locally (ellmer 0.4.2 installed, pkg-r requires >= 0.5.0) — CI will be its first full run.

simonpcouch (approved):

PR #277, opened 2026-09-04
Read a Snowflake catalog

jat255:
Third PR of the catalog import stack (kata xp65). Stacked on #275. Ports the catalog-import half of `catalog-snowflake.R`: the current namespace, `SHOW OBJECTS` listing and exact lookup, `DESC TABLE` column description, and the identifier quoting each needs.

The queries are kept apart from the interpretation of their rows, following the R structure: the row readers are pure and run against real `SHOW OBJECTS` and `DESC TABLE` shapes, and the query builders are checked by recording the SQL they send. 

Two behaviors worth naming. `SHOW` caps its output and a full page cannot be told apart from a page that happened to fill, so a full page is refused rather than silently truncated. And `SHOW OBJECTS LIKE` ignores case, so its reply is filtered by exact name: asking for `ORDERS` and getting back `orders` means a differently-cased table exists, not the one that was named.

## Deliberately out of scope: semantic views

About half of `catalog-snowflake.R` is Snowflake semantic views, and none of it is included in this PR. It is a feature rather than a part of catalog import Tracked locally as kata `gcgj`, which also flags the question of how it relates to commons' own semantic layer.

## Verification

653 tests, ruff and pyrefly clean. Worth knowing: the exact-name filter was originally uncovered. Perturbing it failed nothing, because the only test reaching that path supplied an empty reply. A near-match case covers it now, and that is why the perturbation pass is worth doing rather than trusting a green suite.

### DIFF
--- a/pkg-py/src/commons/_catalog/_snowflake.py
+++ b/pkg-py/src/commons/_catalog/_snowflake.py
@@ -158,6 +158,10 @@ def exact_relation(backend: Any, selector: Selector) -> Relation | None:
         f"SHOW OBJECTS LIKE {_quote_string(selector.table or '')} "
         f"{_namespace_target(Selector(namespace.catalog, namespace.schema))}"
     )
+    # R checks this only when listing a namespace. Checked here too, because
+    # a capped reply that dropped the match reports the table as absent, and
+    # a wrong answer is worse than a loud one.
+    check_show_complete(rows, "relations")
     for relation in relations_from_show(rows):
         if relation.id.table == requested.table:
             return Relation(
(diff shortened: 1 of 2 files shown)


jat255:
Manual review completed 2026-09-05 23:20 MDT by @jat255. Findings: 1 MEDIUM (untested current-namespace fallback in `exact_relation`) and 1 LOW (unpinned schema-only `IN SCHEMA` path) — both closed with tests in 1f35c73. 3 INFO, all documented divergences from `catalog-snowflake.R`, no action needed. Local checks green: ruff, pyrefly, 772 tests. Removing `needs-manual-review`.

PR #278, opened 2026-09-04
Read a Databricks catalog, and pin catalog rows as a shared fixture

jat255:
Fourth PR of the catalog import stack (kata `xp65`). Stacked on #277. Ports the catalog-import half of `catalog-databricks.R`, and moves both readers' row interpretation into a shared fixture.

Metric views are out of scope with the rest of the semantic-model work (kata `gcgj`). 

## Row interpretation is now a shared fixture

Which rows are relations, what kind each is, which comments count as prose, and where a `DESCRIBE` reply stops being columns are all user-observable, so they belong in `tests/shared/` rather than being asserted in each language. `catalog-rows.json` carries the rows both warehouses actually return and what both implementations must make of them. Running the queries stays each language's own business; only their replies are a contract.

Booleans travel as strings, because a bare JSON `true` reads back as a logical in R and an absent one as `NULL`, which would let the two runners disagree about a fixture neither had changed.

## R changes

@simonpcouch 

`R/` has no new behavior, but does have one refactor. Snowflake's column shaping was inline in `snowflake_describe_relation()`, which also runs the query, so the contract it implements could not be reached or tested without a live connection. It is now `snowflake_describe_rows()`, called by that same function with the same rows and returning the same data frame. Databricks already had `databricks_columns_from_describe()` split out, so it needed nothing and this makes Snowflake's implementation more consistent.

In the test suite, two hand-written tests are replaced by fixture runners with their expectations now pinned in `catalog-rows.json`. The Python suite asserts the same cases.

Why it is safe: `test-catalog-snowflake.R` and `test-catalog-databricks.R` both pass.

## Verification

677 Python tests, ruff and pyrefly clean. The reader was confirmed to bite by removing the `#` metadata cutoff and by removing the `hive_metastore` fork.


jat255:
Manual review completed 2026-09-05 23:35 MDT by @jat255. Findings: 1 MEDIUM (untested `hive_metastore` fork in `exact_relation`) and 1 LOW (unpinned backtick doubling in identifier quoting) — both closed with tests in 9f15c41. 4 INFO, all documented or fail-safe divergences from `catalog-databricks.R`, no action needed. Note: the two replaced R tests each dropped a `dbQuoteIdentifier` rendering expectation, so "none was dropped" is slightly off — the substance is pinned by the fixture's `expect_identical` on the `Id`, and the rendering is DBI's contract. Local checks green: ruff, pyrefly, 796 tests. R tests not run locally (ellmer 0.4.2 < 0.5.0 here); CI's R CMD check passed on this head. Removing `needs-manual-review`.

pkg-r/R/catalog-snowflake.R
```diff
@@ -468,6 +468,12 @@ snowflake_describe_relation <- function(con, id, call = rlang::caller_env()) {
       )
     }
   )
+  snowflake_describe_rows(rows)
+}
+
+# Split from the query so the row handling is shared with the Python suite
+# through tests/shared/catalog-rows.json.
```
simonpcouch, line 475:
```suggestion
```

simonpcouch: such a satisfying refactor lol

simonpcouch (approved):

PR #279, opened 2026-09-04
Run each run_r test on its own later loop

jat255:
The R test suite spends about ten minutes doing nothing, and this makes it stop, without touching package code. Each `run_r` test now runs on its own `later` loop, so the timers a call leaves behind are discarded with the test instead of stalling later test files.

> This is a workaround for the issue in #267 to get the tests running more quickly without touching production R code. When there's bandwidth for a proper review, #268 should get merged instead/in addition.

## Summary

`run_r` leaks a 601 second reap timer and a 60 second timeout timer per call, and `shiny::testServer()` reads an output by spinning until `later::loop_empty()`, so the first such read after `test-run-r.R` waits for the farthest one. #267 has the mechanism and the measurements.

`later::later()` and `later::later_fd()` schedule onto `later::current_loop()`, so a test on its own loop cannot leak past itself. `local_temp_loop()` switches the current loop for the caller's scope and restores it on exit; every test in the file calls it first.

## Review Notes

This is containment in the test suite, not a fix. The leak is untouched, #268 cancels the timers, and this branch should be reverted when that one lands.

`local_temp_loop()` calls the unexported `later:::setCurrentRegistryId`, for the reason given in the comment above it. Wrapping each body in the exported `later::with_temp_loop()` also works, but the 80 column limit then reindents all 22 test bodies. A `test_that_local_loop()` wrapper avoids that and was rejected because it breaks testthat's failure locations.

The new global-loop test skips in a full-suite run: `test-chat.R` already leaves callbacks on the global loop, so the test skips when the loop was busy at file load rather than fail for something it did not cause. Its companion, which asserts leftovers land on the loop the call ran on, always runs.

## Testing

The whole suite passes in 50 seconds, against about eleven minutes before, and `test-trajectory-review.R` in the same session goes from 594 seconds to 1.4. `R CMD check` was not run locally, since no package code changed.

Refs #267

## R changes

Nothing under `pkg-r/R/` changed, so the package has no behaviour delta: no function was touched and no call site moved. The diff is one test file.

The helper is a withr-style local: record the current loop, create and enter a fresh one, defer restoring the old one and destroying the new one. Ordering is the part to check. `withr::defer` is LIFO and each test enters its loop before creating a worker, so the worker's deferred `worker_close()` runs while its loop is still alive; swapping those two lines in a test would tear the loop down underneath a closing worker.

Confinement rests on timers being scheduled through `later::current_loop()`, which is what the leak already depends on. I checked it rather than assuming: with the helper on only the tests using `local_worker()` the stall persisted, because several tests build a worker through `test_agent()`, and it went away once all 22 tests had it.

Worth your scrutiny: the `later:::` call is the one thing a `later` upgrade can break, and it would fail loudly rather than quietly resume leaking. No hand-written tests were deleted.

### DIFF
--- a/pkg-r/tests/testthat/test-run-r.R
+++ b/pkg-r/tests/testthat/test-run-r.R
@@ -1,6 +1,30 @@
 # These tests exercise the real worker: they need an installed commons (the
 # worker loads the package from the library, not from load_all()).
 
+# Every run_r call arms later timers that outlive it. Run each test on its own
+# loop so the leftovers die with it: shiny::testServer() reads outputs
+# through shiny:::wait_for_it(), which spins until later::loop_empty(), so a
+# timer left on the global loop stalls other files' testServer() blocks for its
+# full delay (#267). later::with_loop() and later::with_temp_loop() are the
+# exported analogs, but both wrap an expression rather than a scope, so this
+# calls the internal they use to switch loops.
+local_temp_loop <- function(env = parent.frame()) {
+  old <- later::current_loop()
+  loop <- later::create_loop(parent = NULL)
+  later:::setCurrentRegistryId(loop$id)
+  withr::defer(
+    {
+      later:::setCurrentRegistryId(old$id)
+      later::destroy_loop(loop)
+    },
+    envir = env
+  )
+  invisible(loop)
+}
+
+# Only meaningful when nothing else has queued work on the global loop.
+global_loop_clean <- later::loop_empty(later::global_loop())
+
 local_worker <- function(env = parent.frame()) {
   worker <- new_r_worker()
   withr::defer(worker_close(worker), envir = env)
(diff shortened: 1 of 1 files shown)


pkg-r/tests/testthat/test-run-r.R
```diff
@@ -1,6 +1,30 @@
...
+# loop so the leftovers die with it: shiny::testServer() reads outputs
+# through shiny:::wait_for_it(), which spins until later::loop_empty(), so a
+# timer left on the global loop stalls other files' testServer() blocks for its
+# full delay (#267). later::with_loop() and later::with_temp_loop() are the
+# exported analogs, but both wrap an expression rather than a scope, so this
+# calls the internal they use to switch loops.
+local_temp_loop <- function(env = parent.frame()) {
+  old <- later::current_loop()
+  loop <- later::create_loop(parent = NULL)
+  later:::setCurrentRegistryId(loop$id)
```
simonpcouch, line 14: Not big on `:::` here🙃 This will likely introduce an R CMD check NOTE.

If you're game to wait a day, I can have a fix for this in!

jat255:
Closed in favor of #268

PR #281, opened 2026-09-04
Check warehouse session identity and query access

jat255:
## Summary

Fifth PR of the catalog import stack (kata task `xp65`). Stacked on #278. Adds the access half of the catalog security contract to the Python package, and fixes some issues found in the R one during the rewrite.

A warehouse (snowflake, databricks, etc.) decides who may read what, so commons asks with a query that returns no rows and determines if it has access via the failure (or success). If access is not successful, the error is classified: an authorization refusal is stable, so it is cached per relation; a transient failure is retried on the next touch; anything unrecognized is neither. 

Semantic-model probing is the other half of that contract and is not part of this PR (it's included with the semantic model work -- kata task `gcgj`). Nothing in `pkg-py/src` calls the new module yet: R reaches its equivalents from `data_source_connection()` and `source_describe()`, and Python has no construction or first-touch surface until #282 and the retrieval work.

## Shared fixtures

Four behaviors are defined in `tests/shared/`: 
  1. How a driver failure classifies (`catalog-access-errors.json`)
  2. Which parts of a session identity a refusal reports (`catalog-session-changed.json`)
  3. What construction reports when several named relations fail at once (`catalog-access-precedence.json`)
  4. The label a selection entry ends up under (`catalog-relation-labels.json`). 
 
Both the Python and R side test through these fixtures. 

## Verification

860 Python tests, ruff and pyrefly clean. R: `test-catalog-security.R` (142), `test-catalog-snowflake.R` (68), `test-catalog-databricks.R` (60), `test-catalog-merge.R` (65), `test-catalog-search.R` (28), `test-data-source.R` (67), no failures.

## R changes

@simonpcouch 

Five changes on the R side. Three are fixes for defects found by writing the same contract a second time. The fourth resolves a divergence between the two packages in how they report a relation: both now print the fully-qualified `[CATALOG].[SCHEMA].[TABLE]` rather than just the table name. The fifth fixes a regression the fourth caused, found by @simonpcouch running the live warehouse tests.

1. **Access classifier no longer reads its own probe statement.** The probe (`SELECT * FROM <relation> WHERE 1 = 0`) embeds the relation's name in the error text the classifier scanned, so an absent table named `PERMISSION_DENIED_EVENTS` classified as an authorization refusal. The classifier now scans the driver's message only (new `catalog_driver_message()` strips the `<SQL>` tail and prefers `err$parent`). Misreads now land in `unknown`, which denies without caching — safe direction, but it changes which failures are remembered. (this is an edge case for sure, but good to be technically correct).
2. **Exact selections are labelled with the resolved (qualified) id.** `tables = "orders"` now shows as `ANALYTICS.PUBLIC.ORDERS` in listings, search scoring, and errors, matching the Python implementation so they can share a test. Opt-in per caller via a new `requested` argument defaulting to the old behavior. The resolved id is also what the access probe and every later query name the relation by, not only what the listing displays; the warehouse's own spelling is kept alongside it as `identity`. Entries the warehouse never reported keep their authored name.
3. **Session refusal names only the fields that differ.** Previously it listed every compared field, telling Databricks users a role changed on a backend with no roles. Message text only; condition class unchanged. Docs now scope the role snapshot to Snowflake.
4. **`catalog_require_queryable_relations()` reports all problems at once** instead of raising on the first unreadable relation and discarding the missing names it had collected. Failure path only; successful constructions are unchanged.
5. **A Databricks temporary view is no longer given a namespace it does not have.** `SHOW TABLES` lists the session's temporary views alongside the schema's own tables, and `databricks_list_hive_relations()` stamped the selected namespace onto every row. Such a view belongs to no schema and answers only to its bare name, so the id it was given named nothing. Change 2 made that id reach the probe, which is what broke `live Databricks handles quoted relation and column names`. The listing now leaves those rows bare, and `catalog_match_exact_relation()` adopts the resolved label only when the warehouse reported a namespace of its own. This also fixes the namespace-selection path, which produced unqueryable ids before this PR with nothing covering it.

**Blast radius.** The exact-match and registry changes are reached only from the two backend implementations; `catalog_require_queryable_relations()` only from `data_source_connection()`. The hive listing serves both the exact lookup and a `hive_metastore` namespace selection, and no other backend reaches it. The classifier and session check sit on every probe and query path. Only `catalog_match_exact_relation()` changed signature, and the new argument defaults to the old behavior.

**Why it is safe.** All six R test files pass, and each fix was verified by reverting it and watching the new tests fail. The classifier change is pinned by three fixture cases where the relation name contradicts the driver's message. The temporary-view case is pinned by a fifth case in `catalog-relation-labels.json`, which now names the backend each case runs against; the live Databricks test needs credentials I do not have, so that run is still worth repeating.


jat255:
Manual review completed 2026-09-06 13:16 MDT.

Reviewed across correctness, public API and cross-implementation parity, test adequacy, docs, and security. One high-severity finding, three medium and eight low, all fixed in 2af2fd4; roborev job 440 passed with no findings on the fix. The R-side changes are written up in the description.

The finding worth recording: the access classifier matched against the whole exception text, which on both backends contains the probe statement, so a relation's own name could decide its classification. An absent `PERMISSION_DENIED_EVENTS` read as an authorization refusal and was cached as one for the life of the source.

simonpcouch (changes requested):
With this PR, I see the following failure in `live-warehouses`:

```text
Error in `data_source(con, tables = table)`: `tables` names table not on the
connection: "hive_metastore.default.commons quoted.table".
```

jat255:
> With this PR, I see the following failure in `live-warehouses`:
> 
> ```
> Error in `data_source(con, tables = table)`: `tables` names table not on the
> connection: "hive_metastore.default.commons quoted.table".
> ```

## Root cause

`databricks_list_hive_relations()` (`pkg-r/R/catalog-databricks.R:404`) stamps the selected namespace onto every row of `SHOW TABLES`:

```r
lapply(rows$tablename, function(table) {
  list(id = DBI::Id(catalog = components[["catalog"]], schema = components[["schema"]], table = table), ...)
})
```

`SHOW TABLES` also returns session-scoped temporary views, with an empty `database` and `isTemporary = TRUE`. Those objects have no namespace: they are addressable only by their bare name. The listing fabricates one for them, so the constructed id `hive_metastore.default.\`commons quoted.table\`` names nothing.

On `main` that was harmless for an exact selection, because `catalog_match_exact_relation()` overwrote `relation$id` with the authored bare id, and `registry$ids` (which is what `table_ids` and every later query quote, `pkg-r/R/data-source.R:212`, `pkg-r/R/pool.R:68`, `pkg-r/R/tools.R:933`) carried the bare, addressable name. This PR changes that line to `relation$id <- requested`, so the fabricated namespace now reaches the probe and every subsequent query, and the probe fails.

Two things worth saying about the blast radius:

- The PR description says "the warehouse-facing `identity` id is unchanged". That is true of the `identity` field, but `identity` is not what queries use; `registry$ids` is, and it now carries the qualified requested name. So this change is not label-only.
- The underlying listing defect predates the PR. A *namespace* selection over `hive_metastore` (`tables = DBI::Id(catalog = "hive_metastore", schema = "default")`) already produces unqueryable ids for temp views on `main`; nothing covers that, so it has gone unnoticed. The label change just removed the accident that was masking it on the exact-selection path.

jat255:
Fixed in fbdfd59.

`SHOW TABLES` lists the session's temporary views next to the schema's own tables, and the hive listing gave every row the selected namespace. A temporary view belongs to no schema and answers only to its bare name, so `hive_metastore.default.commons quoted.table` named nothing. Change 2 in the description made that name reach the access probe, which is what you hit.

The listing now leaves such a row bare, and an exact match takes the resolved label only when the warehouse reported a namespace of its own. A new case in `catalog-relation-labels.json` pins it. This also fixes a `hive_metastore` namespace selection, which built the same unqueryable ids before this PR.

simonpcouch (approved):

PR #282, opened 2026-09-04
Import a warehouse catalog when a data source is built

jat255:
## Summary

Sixth PR of the catalog import stack (kata task `xp65`). Stacked on #281. Python only.

Everything this catalog stack has been foundational with nothing calling it. This PR wires it in: a Snowflake or Databricks engine now imports its catalog when the source is built. The session identity is read first; relations named in `tables` are access-checked before anything is described, so a name the caller got wrong fails at construction rather than mid-conversation; the listing is folded into the authored dictionary; and the session is read again at the end, because a role that moved during discovery invalidates what was just learned.

A selection defaults to the namespace the connection already points at. A namespace has to be named with a `Selector`, because `ANALYTICS.PUBLIC` on its own does not say whether `PUBLIC` is a schema or a table; a plain string is read as a relation, the way it is everywhere else. `exclude` is accepted only for a warehouse, since nothing else has a listing to drop from.

The merge re-keys the dictionary to the warehouse's labels, so a table's definitions are now looked up under the name the author gave it. Definitions over a column the warehouse spells differently are refused for now: the compiler cannot yet bind an authored name to the discovered one, and lowering them as written would emit SQL against columns that do not exist. The next PR in the stack replaces that refusal with real binding.

`dictionary` is taken by `from_engine()` and `from_board()` because the merge needs it during construction. Its definitions are lowered at the end of every constructor rather than in `data_source()`, so a source built through a constructor directly carries them instead of silently dropping them.

The manifest starts with every relation unknown, including the ones construction just probed. Carrying that answer forward would save a round trip on first touch at the cost of serving a grant revoked in between, and the R implementation re-probes for the same reason.

## Verification

898 Python tests, ruff and pyrefly clean. Fake Snowflake and Databricks backends stand in for the network, taking their row shapes from `catalog-rows.json` rather than restating them, so the real readers, registry, merge, and access code all run.

The order the import runs in is what makes the source safe, so it is asserted as a sequence of steps read off the query log rather than left implicit. Three inversions were each confirmed to fail exactly one test: reading the session identity after the registry, probing named relations after the merge, and moving the closing session re-read ahead of the merge. Databricks was confirmed to be genuinely exercised by wiring it to the Snowflake reader and watching three tests fail.

## Known limitations

`EngineBackend.query()` opens a connection per call, so the session snapshot, catalog listing, access probes, and the user's query can run on different pooled connections. The warehouse still enforces authorization on every statement, so the cost is a stale or misleading error rather than an access bypass. Pinning a connection for the source's lifetime changes the backend contract and needs an R counterpart, so it is left as a known limitation.

A selection entry is matched against the listing case-sensitively, in both packages: `SHOW OBJECTS LIKE` ignores case, so accepting a near match would bind the selection to a differently-cased table rather than the one named. `tables = ["orders"]` therefore does not find `ORDERS`. Treating an authored name as an unquoted identifier is a change to the shared contract and is not made here.

Construction probes one relation at a time, and a failed session-identity read is reported as an opaque error rather than classified as transient. Both match the R implementation, so neither is changed unilaterally.

### DIFF
--- a/pkg-py/src/commons/_catalog/_core.py
+++ b/pkg-py/src/commons/_catalog/_core.py
@@ -76,6 +76,11 @@ class TableRegistry:
     relations: dict[str, Relation]
     validate: dict[str, TableId]
     namespace_selected: bool
+    # The relations the listing reported and exclude then removed, so a
+    # caller left with nothing, or with a dictionary entry that now matches
+    # nothing, can say which it was. A name the warehouse never had is not in
+    # here: exclude is not what made it absent.
+    dropped: list[Relation] = field(default_factory=list)
 
 
 @dataclass
@@ -221,6 +226,9 @@ def table_registry(
         relations.extend(list_relations(selector))
 
     keep = excluded([item.name for item in relations], exclude)
+    dropped = [
+        item for item, hidden in zip(relations, keep) if hidden and item.discovered
+    ]
     relations = [item for item, hidden in zip(relations, keep) if not hidden]
     validate = [
         table_id
@@ -252,6 +260,7 @@ def table_registry(
         relations=labelled,
         validate={table_id.label: table_id for table_id in validate},
         namespace_selected=namespace_selected,
+        dropped=dropped,
     )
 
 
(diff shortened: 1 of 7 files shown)


jat255:
Manual review completed 2026-09-06 14:18 MDT.

Reviewed across correctness, public API and cross-implementation parity, test adequacy, and docs. Three high-severity findings and twelve medium, fixed in 2624f95; roborev needed three further rounds on that commit and now passes clean (jobs 441, 443, 445, 447, each closed with its disposition).

The two findings worth recording. The construction order is what makes a warehouse source safe, and nothing asserted it: three separate inversions each passed the whole suite, including moving the closing session re-read ahead of the merge, which is the one thing it exists to cover. And only Snowflake was ever exercised; the fake could not answer a Databricks query at all, so wiring `databricks` to the Snowflake reader passed everything.

Three findings were declined with reasons rather than fixed, and the two behaviours that match the R implementation are now named in the description as known limitations rather than left implicit.

PR #283, opened 2026-09-04
Bind definitions to the names the warehouse reported

jat255:
Seventh (and last) PR of the catalog import stack (kata `xp65`). Stacked on #282. Python only.

## Summary

A definition's expression names the columns the author wrote. A catalog import re-keys the dictionary to what the warehouse reported, which on Snowflake means upper case, so the expression has to be rewritten before it is lowered or the SQL would name columns the warehouse does not have. This is the catalog column binding that f6hz deferred to xp65.

Column paths in the IR and in a `COLUMNS(...)` selection take the discovered spelling. A reference to a sibling definition is a name in the dictionary rather than a column in the warehouse, so it is left for composition to inline. An authored column the relation does not have is refused rather than emitted, and so is an authored table with definitions that matched no selected relation, since dropping it would leave the agent told about a governed definition that is not there.

How strictly a `COLUMNS(...)` selection binds depends on how it was written. One the author wrote out by name refuses a column the relation does not have, because naming a column that is not there is worth reporting. One resolved from a pattern or from `*` drops it instead: the author named no column, the selection is resolved over the authored columns, and refusing would take every wildcard definition on the table down over one stale dictionary entry. That already matched how the other side of the intersection behaves, since a warehouse column the dictionary does not describe was never selected. A derived selection left with nothing still refuses.

`DefinitionExport.columns`, the list of columns a definition reads, keeps the authored spelling. It describes the dictionary the author wrote rather than the relation it matched, so it is the one place authored names survive binding, and both packages report it that way.

This replaces the temporary refusal the previous PR shipped.

## Verification

908 Python tests, ruff and pyrefly clean. Binding is exercised through the public constructor against a warehouse that renames the column, not only through the compiler directly, so the argument reaching `attach_compiled_definitions()` is covered rather than assumed.

## Not here

Both packages now implement this binding, so the behaviour belongs in `tests/shared/` rather than in the Python suite alone. That is kata `5hfw`, kept out of this PR to leave it Python-only. Two divergences found while reviewing this change are recorded there: R binds every selection column alike, so it has no derived-versus-named split, and the two packages name a different table in the refusal. R also threads a `strict` flag through binding that Python does not carry; they agree today only because R's flag is unconditionally true at its one call site.

### DIFF
--- a/pkg-py/src/commons/_catalog/_core.py
+++ b/pkg-py/src/commons/_catalog/_core.py
@@ -87,6 +87,10 @@ class TableRegistry:
 class MergedDictionary:
     dictionary: Any
     relations: dict[str, Relation]
+    # What the merge matched, for the definition compiler: "tables" maps each
+    # authored table name to the relation label it matched, or None, and
+    # "columns" maps each authored column name to the spelling the warehouse
+    # reported. None when there was no catalog to match against.
     definition_bindings: dict[str, Any] | None
 
 
--- a/pkg-py/src/commons/_catalog/_core.py
+++ b/pkg-py/src/commons/_catalog/_core.py
@@ -402,11 +402,7 @@ def merge_dictionary(
     return MergedDictionary(
         dictionary=dictionary,
         relations=relations,
-        definition_bindings={
-            "tables": matches,
-            "columns": column_matches,
-            "strict": True,
-        },
+        definition_bindings={"tables": matches, "columns": column_matches},
     )
 
 
--- a/pkg-py/src/commons/_catalog/_import.py
+++ b/pkg-py/src/commons/_catalog/_import.py
@@ -17,11 +17,9 @@ from .._data_source import TableId
 from . import _databricks, _snowflake
 from ._core import (
     Manifest,
-    MergedDictionary,
     Relation,
     Selector,
     check_exclude,
-    has_suffix,
     id_type,
     merge_dictionary,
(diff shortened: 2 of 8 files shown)


jat255:
Manual review completed 2026-09-06 14:09 MDT. Reviewed the stacked diff against #282 in a local worktree; ruff, pyrefly, and the full pytest suite (887 passed) all green. No blocking findings. Removing `needs-manual-review`.

jat255:
Manual review completed 2026-09-06 16:58 MDT.

Reviewed across correctness, public API and cross-implementation parity, test adequacy, and docs. The parity pass verified the central claim function by function against `definition_bind_one()`, and mutation testing killed every one of the six core behaviours, so the binding tests are real rather than decorative. Five medium findings fixed in a36941d; roborev passes clean (jobs 448 and 449, both closed with their disposition).

The finding worth recording: a `COLUMNS(...)` selection resolved from a pattern or from `*` was bound as strictly as one written out by name, so a dictionary that still documented a column the warehouse had dropped took every wildcard definition on that table with it, and named a column that appears nowhere in the expression. Derived selections now drop what the relation lacks, which is what the other side of that intersection already did.

Pinning this behaviour as a shared fixture is deliberately not here, to keep the PR Python-only: it is kata 5hfw, which now also carries two divergences this review found, the derived-versus-named split and the table each package names in the refusal.

PR #284, opened 2026-09-04
Python context layer: pin dictionary prose as retrievable chunks

jat255:
Third of four PRs for M4, the Python context layer (kata `vp7e`). Stacked on [#241](https://github.com/posit-dev/commons/pull/241); [#285](https://github.com/posit-dev/commons/pull/285) follows this and wires the chunks into an agent's layer on top.

This PR adds shared fixtures defining the expected way prose should be transformed into retrieval chunks. `tests/shared/context_layer.json` gains a `dictionary_context_chunks` section and both suites run it against their own reader: dataset details, then tables in declaration order, then glossary terms, then governed definitions, with a table that has no prose contributing nothing and column-level content staying out.

Implementing this PR found that python needed a guard for a source carrying no dictionary at all.

Governed definitions get no fixture case, deliberately. Both packages render the chunk text the same way and `definitions.json` pins it, but they compile definitions at different constructors, so they disagree on whether a dictionary not yet attached to a source has any. A case here would pin that disagreement rather than the contract.

## R changes

@simonpcouch 

One functional change in `pkg-r/R/context-layer.R`, required by an error found by the new fixture: `dictionary_context_chunks()` returned a named character vector and now returns an unnamed one.

The names came from `vapply(names(dictionary$tables), ...)`, which labels each chunk with its table name, and `c()` propagated them to the whole vector (with `""` for the dataset details and glossary entries). Those names rode into `augment_context_layer()` and onto the layer's `docs` vector. Nothing ever read them, but the R side was carrying metadata Python can't represent, which made the shared fixture unsatisfiable for both suites. The fix is a single `unname()`; chunk text and order are unchanged.

Blast radius: two call sites, neither reads names. `augment_context_layer()` concatenates the vector into documents, and the one test caller (`test-definitions.R:401`) uses `grepl()` and positional subsetting.

Verified by running the four test files that touch this path: `test-context-layer.R` (65 passing), `test-data-dictionary.R` (71 passing, 3 skipped on CRAN), `test-definitions.R` (114 passing), `test-citations.R` (84 passing). I skipped the full R suite — it hangs ~10 minutes on `run_r`'s uncancelled `later` callbacks, fixed in #268.

One thing worth scrutiny: the new runner in `test-context-layer.R` feeds each fixture case through the real `new_data_dictionary()` reader, which works because both languages parse the same authored `data-dict.yaml`. If that ever diverges, this runner is where it will surface.

No hand-written tests were deleted; the new test is additive. The generated fixture `pkg-r/tests/testthat/fixtures/shared/context_layer.json` comes from `scripts/sync-shared-fixtures.sh` and needs no review.


simonpcouch (approved):
Live warehouse tests pass. :)

PR #289, opened 2026-09-05
Show "information" icon (leading to modal) inline

### LINKED ISSUE #264: Put provenance modal icon inline
I think the icon would look better if it was inline with the text

Image

simonpcouch:
Closes #264. Now:

> provenance-icon-inline


simonpcouch:
With more context:

Screenshot 2026-09-05 at 10 23 35 AM

PR #293, opened 2026-09-05
clarify provenance marker language

### LINKED ISSUE #195: Reconsider "defined by your data team" in verified marker language
Image

I'm wondering if some people will think 'what data team,' if they don't have a data team or just made an app for themselves.

simonpcouch:
Closes #195. Rewords the  text in the popups.

| Outcome | Before | After |
|---|---|---|
| Verified answer | This answer comes from a trusted calculation defined by your data team. | This answer comes from a trusted calculation. |
| Cited | This answer includes supporting text verified against a trusted source. | This answer cites context from a trusted source that supports its approach. |
| Untrusted | This answer was not produced by a trusted calculation and has no verified supporting citation. AI can be wrong. | This answer was not produced by a trusted calculation and does not cite trusted context. |

They're purposefully not aligned with the descriptions inside of the modal, which are intended to be consistent with these but phrased in a different way that, hopefully, in combination, will explain the intent.

### DIFF
--- a/pkg-py/src/commons/_provenance.py
+++ b/pkg-py/src/commons/_provenance.py
@@ -44,23 +44,24 @@ PROVENANCE_DISPLAY: dict[Tag, ProvenanceDisplay] = {
     Tag.A: ProvenanceDisplay(
         label="Verified answer",
         icon="trusted-icon.svg",
-        body=(
-            "This answer comes from a governed calculation defined by your data team."
-        ),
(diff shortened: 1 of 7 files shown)


simonpcouch:
cc @jat255, going to go ahead and merge these changes, which do touch the Python side. These slot into infrastructure you've put together, just changing string values—apologies if you would have preferred you had a chance to review these yourself.

PR #296, opened 2026-09-05
remove `MIGRATING.md`

simonpcouch:
Thanks for putting this together @jat255! I don't believe we have any more active PRs that would require this guidance, so I think we can go ahead and remove it.


jat255:
LGTM

PR #304, opened 2026-09-07
feat: single source for the system prompt, rendered by both packages

jat255:
This PR (the first of M5) adds the system prompt to the package sharing mechanism. Python gains prompt rendering, which it did not have before.

## Summary

`prompts/` at the repo root holds `system-prompt.md` and `citation-request.md`, the second of which was previously a string literal in R's `citation_reminder_text()`. `scripts/sync-shared.sh` generalizes the fixture sync to a table of source and destination pairs and copies `prompts/` into both packages; `verify-shared-synced.yaml` replaces `verify-shared-fixtures.yaml` and fails on a stale copy. Because the copies sit inside each package directory, editing a prompt now also triggers R-CMD-check and py-check.

The template language for the prompt is Jinja2: this is directly interpretable using `jinja2` in Python, and this PR adds a small renderer in R to handle the subset of jinja the template uses. `glue` could not be used as the shared syntax because its conditionals are R expressions, so a Python renderer would have to evaluate R. Jinja2's conditionals are data, and a subset this small is cheap to implement in R and to pin with fixtures.

## Review notes

The five expected prompts in the test fixtures are the exact text the R implementation produced before this change, so the PR is effectively behavior neutral on the R side.

Two renderers of one template is a standing divergence risk, so both implementations have a test to reject anything outside the subset instead of letting a template render in Python and fail in R. 

Out of scope, as planned: Python's `system_prompt_data()` and the fields that need a data source. The following files are generated from the shared data and don't need a review:: `pkg-r/inst/prompts/`, `pkg-py/src/commons/prompts/`, and `pkg-r/tests/testthat/fixtures/shared/`.

## Testing

Full R suite: 7023 pass, 0 fail. Python: 920 pass, ruff and pyrefly clean. The built wheel contains `commons/prompts/`.

## R-side summary

@simonpcouch 

`commons_system_prompt()` keeps its signature and returns the same text as before for every input the package can build. Under it, `render_system_prompt()` uses a new internal `render_template()` instead of glue, `system_prompt_data()` trades the two rendered tool-output lists for one flag per tool, and `citation_reminder_text()` reads its words from the generated `inst/prompts/` copy.

The only reachable behaviour change is `citation_trust_exception()`, which names trusted tools in a fixed order rather than registration order. That differs only for a hand-passed tool list.

All of this is internal: `system_prompt_data()` and `render_system_prompt()` are reached only through `commons_system_prompt()`, itself called only by `Commons$set_system_prompt()`; both call sites of `citation_reminder_text()` get the same words; the two deleted `*_tool_output_text()` helpers had no other callers.

The expected prompts in the shared fixture came from R as it stood before this change, and R still reproduces them exactly, with the full suite passing.


pkg-r/R/citations.R
```diff
@@ -187,66 +177,29 @@ available_tool_names <- function(tools) {
...
-    if ("run_sql" %in% tools) "- Query result rows from `run_sql`.",
-    if ("run_r" %in% tools) {
-      "- Code, measure source, plots, and textual output from `run_r`."
-    }
-  )
-  paste(items, collapse = "\n")
+# Which outputs are citable depends on which tools the agent registered, so the
+# template branches on one flag per tool. The order here is the order the
+# citation sections read in, not the order tools are registered.
+TRUSTED_TOOLS <- c(
```
simonpcouch: Snake case, please!

pkg-r/R/prompt.R
```diff
@@ -60,15 +70,254 @@ render_system_prompt <- function(
...
+# glue. This renders the subset the template uses: `{{ name }}` substitution,
+# `{% if name %}` / `{% if not name %}` / `{% else %}` / `{% endif %}` on plain
+# names, and `{% raw %}` for literal braces. Python renders the same file with
+# jinja2 configured for the same whitespace handling; tests/shared/
+# prompt-render.json pins the two to the same output.
+#
+# Substituted values are inserted literally and never rendered again, which is
+# what lets governed-definition tokens and app-authored instructions contain
+# template syntax of their own.
+TEMPLATE_TOKEN <- "(?s)\\{\\{\\s*[A-Za-z_][A-Za-z0-9_]*\\s*\\}\\}|\\{%.*?%\\}"
```
simonpcouch: Same here.

simonpcouch (approved):
Nice, thanks! Warehouse tests pass.

pkg-r/R/prompt.R
jat255: Done in de92f9d — renamed to `template_token`.

pkg-r/R/citations.R
jat255: Done in de92f9d — renamed to `trusted_tools` (and `cited_tools` below it).

PR #317, opened 2026-09-08
feat(py): add agent's tools, prompt, and chat client to public API & fix describe_table bug

jat255:
Closes kata `y2vs`.

`Commons` composes a chatlas `Chat` object only forwarded certain methods, which didn't include the tools it registered or the prompt it rendered. R gets both by default by inheriting directly from `ellmer::Chat`, so the only Python route was accessing the the private `._client` attribute. This PR adds [`get_tools()` and a read-only `system_prompt`, plus `client`](https://github.com/posit-dev/commons/pull/317/files#diff-73bffe38e0d176e55662779d49a175bfe7797ad9fc85c1313f38b582b7ee04f8) as the documented escape hatch to the rest of chatlas.

[A new test](https://github.com/posit-dev/commons/pull/317/files#diff-bcb6dbee60de8cd3ed1f99b6ec2dc4b3fb7b5515b940e83604f7d484a397fddb) pins that `Commons` still carries no `conversation_id`. D5 depends on shinychat's `hasattr` check failing so M8's chat adapter owns that property, and a convenience property added later would break it silently. No `tests/shared/` fixture: R reaches these by inheritance and Python by forwarding, so there is no single observable behaviour to pin without pinning each language's call syntax, which is implementation detail by this repository's rule.

Also adds [`demo.py`](https://github.com/posit-dev/commons/pull/317/files#diff-6da12276f0ec8796b07df60b8b267a44259310c039d54f8a3602eb5f9f59b0f2) and [`demo.ipynb`](https://github.com/posit-dev/commons/pull/317/files#diff-96f9580b8ca16f19b463b3792449a26090ee6c7dadf9a9e20122940f0ab531f6), the counterpart to `pkg-r/inst/demo.R`. The gap addressed by this PR was found while building the demo , and the demo reads the tools and the prompt through the new accessors. 

## Small bug fix (affects R)

Building the demo also exposed a defect worth fixing: `describe_table` summarizes at most five sampled rows but presented that summary as the table's entire contents, so a column's value list read as its full domain. Asking how many acres are ponderosa pine confidently returned that the table holds none, naming exactly the three forest types in the first five rows of eight. The heading of `describe_table` now scopes what follows, and its tool description says what that summary cannot establish. Both sit on the tool, next to the output they describe, and nothing in the system prompt changes.

This is a mitigation rather than a repair, and #319 holds the defect underneath: the summary is computed over five sampled rows while carrying the shape of a whole-table profile. Measured over 20 runs per condition against `claude-sonnet-5`, the tool changes answer that question correctly 17 times in 20. The same sentence in the system prompt reached 65/65 and is left out on review, since prompt text there would describe a tool result the model may not have seen. A true row count is also left out: it would cost a `COUNT(*)` per call, unbounded on a large table or view.

Verified against a live model rather than a scripted client, `us.anthropic.claude-sonnet-5` on Bedrock: the script answers all three of `demo.R`'s questions, all 26 notebook cells run clean, and the ponderosa pine question now comes back verified from the measure. From `pkg-py`: ruff clean, pyrefly reports 0 errors over `src tests`, and 1311 tests pass.

## R-side summary

@simonpcouch 

One line of hand-written R changes: the heading `describe_table_tool()` puts above its sample block, in [`R/tools.R`](https://github.com/posit-dev/commons/pull/317/files#diff-c78651d41d6dc764b557809df924b62b89e0acc1252e654da8b767c8dc18630d). The tool description gains the same two sentences, in both of its branches. Everything else under `pkg-r/` is `scripts/sync-shared.sh` output: the shared fixture copies under `tests/testthat/fixtures/shared/`.

The behaviour delta is that every `describe_table` call on a table now introduces its sample with `Sample summary (the sampled rows only, not necessarily every row):` instead of `Sample summary:`, and its tool description carries two more sentences. The rendered system prompt does not change, and nothing else about the tool's value moves: the schema, the dictionary entry, the definitions, the first-touch context, and what counts as citable are all unchanged. The direction is strictly more scoping, never less.

Blast radius is one call site. `describe_table_tool()` is called only from the tool registration in `R/tools.R`, so the change reaches the `describe_table` tool's value and nothing else. `source_describe()` is untouched, and `source_describe_semantic_model()` builds no sample block at all, so a describe of a semantic model is unaffected.


pkg-r/R/tools.R
simonpcouch, line 1: Could we test with only this change rather than with this change _and_ the change to the system prompt? I think this alone should be sufficient to clarify that, and I think it's easier to describe the effect here rather than in the system prompt, where we sort of need to preemptively describe the tool result

jat255: Tested exactly that, 20 runs per condition against `claude-sonnet-5`, on the question that found the bug (how many acres are ponderosa pine, where the two ponderosa stands sit after the first five rows):

- neither change: right 0/5, and no run ran a query
- this change alone: right 16/20
- this change plus the prompt sentence: right 20/20, and every run ran a query

So the heading gets most of the way there but leaves about one in five answers confidently wrong. The failures are all the same shape: `describe_table`, then "the stands table only has three forest types", with no query run. The one that convinced me reported that the values came from "a 5-row sample" and still concluded ponderosa pine was absent, so the model can read the scoping, restate it correctly, and generalise anyway. Fisher's exact on the pooled runs (5 failures in 26 without the sentence, 0 in 25 with it) gives p = 0.05, so this is a strong point estimate rather than a significant result, on one model and one question.

Your placement point still holds and I would rather satisfy it than argue it, because the prompt sentence does have to describe a tool result the model may not have seen yet. Let me try moving the whole message into the tool result, saying what the summary does not establish rather than only labelling it a sample, and test that on its own the same way. If it reaches 20/20 the prompt change comes out and everything lives next to the thing it describes.

jat255: Tried it, and it helps but does not close the gap. Same setup, 20 runs each:

- tool result alone, naming the consequence rather than just the scope (`Sample summary (the sampled rows only; a value missing here is not absent from the table, so query before saying it is):`): right 18/20
- the pair as it stands: right 20/20

Better than the plain scoped heading's 16/20, so your read on where the guidance belongs is right about which wording carries the weight. But two runs still answered that the table has no ponderosa pine, and 18/20 against 20/20 cannot be separated at this sample size (p = 0.49). Pooling every tool-result-only run against every run with the prompt sentence is what has signal: 6 failures in 40 versus 0 in 45, p = 0.009.

The failures are all the same shape, and they are why I do not think a longer heading gets there: the model reads `describe_table`, restates the three types it saw, and answers without querying. One earlier failure quoted "a 5-row sample" back and still concluded absence. Once it decides it already knows the answer, it never re-reads the caveat sitting above the values it is looking at, whereas the prompt sentence lands before it forms that conclusion.

So I have kept both, and left the heading as the shorter scoped version rather than the longer one, since there is no evidence the extra clause adds anything once the prompt sentence is there, and it would sit on every `describe_table` result. Happy to switch to the longer wording and drop the prompt line if you would rather trade about a 15% rate of a confidently wrong answer for keeping everything local. That is a call about the shape of the API rather than about the data, so it is yours to make.

jat255: Moved it onto `describe_table`'s description, in both R branches and Python, and rewrote the sentence. It now reads:

> The sample summary covers the sampled rows only, so it cannot show which values the whole table holds. Before you report that a value is absent, query for it.

Two sentences instead of one, no semicolon, and it states what the summary cannot establish rather than making the reader work through "a value that summary does not list is not evidence the table lacks that value". You were right about that, and the rewrite is in all the places the text appears.

I also tried the following wording:

> The sample summary only contains a subset of rows from the table, so it cannot be used to determine what values the whole table holds. Before you report a value is absent, query in the whole table for it.

but measuring it the same way failed 2/40 times, compared to 0/65 for the first version, so I left it as what Claude generated in the first version.

## Prompt placement

I ran some tests (20 runs per condition) putting the prompt in different places:

- scoped heading alone: succeeded 16/20
- heading naming the consequence rather than the scope: succeeded 18/20
- heading plus the tool description, prompt untouched: succeeded 17/20
- any of those plus the prompt sentence: succeeded 65/65

It appears the failure mode is that once the model decides it knows the answer it does not go back and re-read the caveat sitting above the values it just read, whereas the higher-level prompt arrives before that conclusion forms (this is totally non-deterministic, so that's just a theory).

So I have kept all three for now: the heading, your tool description, and the prompt.

jat255: Okay, after some offline back and forth, this is landing as in 9b7c87d. The guidance now lives only on the tool: the sample-summary heading and `describe_table`'s description, in both R branches and Python. The system-prompt changes are gone, since we can't be sure that won't have other unexpected downstream effects.

For reference, putting the instructions just in the tool call is not guaranteed to return the correct result. In testing, the tool-only version answers the ponderosa pine question correctly 17 times in 20, against 65/65 for the arrangement that also carried the prompt sentence. 

The true underlying design problem is that the model thinks it has access to the entire dataset, when it really only has seen a brief summary. I opened #319 to track how we want to properly fix this going forward.

simonpcouch (approved):
Thanks!

PR #320, opened 2026-09-08
fix(ci): name the deploy target explicitly on both Connect deploys

### LINKED ISSUE #318: deploy job fails: rsconnect-python 1.31.0 rejects the staging deploy
The `deploy` job fails on every pull request and will fail on the next `main` push that touches `pkg-r/**`. The "Deploy to Connect (staging)" step stops with:

> Error: This directory has been deployed before, but the server it was last deployed to, \*\*\*, is no longer saved, so the target has to be named. Pass -n/--name, -s/--server, or --connect-cloud with -A/--account, to say where to deploy.

The "Deploy to Connect (dev)" step succeeds. Only staging fails.

## Cause

`rsconnect-python` 1.31.0 was published to PyPI at 2026-09-08 19:49 UTC. `posit-cli` 0.1.1 requires `rsconnect-python>=1.30,<2`, and `posit-dev/connect-actions/deploy` installs it fresh on every run, so the new version arrived with no change in this repository.

The workflow runs two deploy steps in one job, and both use `path: pkg-r/inst`. The dev step deploys and saves a deployment record into that directory ("Saving deployed information..."), then it removes the saved server nickname when it exits. The staging step then reads that record, which names a server that is no longer saved. Earlier versions of rsconnect-python accepted this. 1.31.0 refuses and asks for an explicit target.

The workflow has always depended on rsconnect-python being lenient here. The new release removed that leniency.

## Timeline

| Time (UTC) | Event | Result |
|---|---|---|
| 18:42 | deploy run ([logs](https://github.com/posit-dev/commons/actions/runs/34264674312)) | pass |
| 19:28 | deploy run | pass |
| 19:49 | rsconnect-python 1.31.0 published | |
| 20:09 | deploy run ([logs](https://github.com/posit-dev/commons/actions/runs/34273119846)) | fail |
| 20:15 | re-run of the failed job | fail |
| 20:46 | deploy run ([logs](https://github.com/posit-dev/commons/actions/runs/34276712859)) | fail |

## What this is not

- Not a code change in this repository. The failing tree differs from a passing tree only in prompt and tool-description text.
- Not the action. `posit-dev/connect-actions@main` is unpinned, but its last commit is 2026-08-28.
- Not the CLI. Both the passing and failing runs installed `posit-cli==0.1.1`.
- Not the secrets. `CONNECT_SERVER_DEV` and `CONNECT_SERVER_STAGING` were last updated 2026-08-05.
- Not the workflow structure on its own. In the passing run the dev step also saved a record, and staging still succeeded.

One thing is inference rather than fact: I did not read the 1.31.0 diff, because its changelog was not reachable through the API. The timing and the error text point at that release. To confirm it, pin `rsconnect-python==1.30.0` and run the job again.

## Options

1. Pin `rsconnect-python` to 1.30.x for the deploy job. This is the fastest way to get CI green, and it confirms the diagnosis.
2. Give the staging step an explicit target with `-s` or `-n` through `rsconnect-args`. The step already passes `content-guid`, so the target is known.
3. Give each deploy step its own copy of `pkg-r/inst`, so the two steps do not share a deployment record.

Option 1 or 2 unblocks CI. The durable fix belongs upstream in `posit-dev/connect-actions`, which can constrain the dependency or name the server for each deploy.

Found while working on #317, whose other 12 checks pass.

jat255:
Fixes #318. The `deploy` job has failed on every run since rsconnect-python 1.31.0 was published.

Both steps deploy `pkg-r/inst`. For a manifest deploy rsconnect derives its deployment-record path from the deploy target, giving `manifest.json/rsconnect-python/manifest.json.json`, which can never be created because `manifest.json` is a file. The save falls back to the user config directory, which the two steps share. The dev step's record names the dev server, whose credential is removed when that step exits, so staging infers a target it can no longer resolve. Earlier rsconnect versions deployed anyway; 1.31.0 asks for an explicit target instead.

[Both steps](https://github.com/posit-dev/commons/pull/320/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3) now pass `--name connect-actions`, the nickname the action logs in under, which skips record inference. The dev step gets it too: it is only safe today because it runs first. `--server` would also name a target, but a trailing slash in the secret would not match the saved credential and would lose its API key.

This couples the workflow to the action's login nickname. The durable fix belongs in `posit-dev/connect-actions`, which could name the server for each deploy itself.

Verified two ways. Locally, a fixture holding a dev record with only the staging credential saved reproduces the error text exactly, and adding `--name` gets past target resolution. In CI, this PR touches `.github/workflows/deploy.yml`, a trigger path, so the `deploy` job runs here.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,6 +18,12 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      # Both steps deploy the same directory, and rsconnect keeps a deployment
+      # record for it in the user config directory, which the two steps share.
+      # The dev step's record names a server whose credential is dropped when
+      # that step exits, so staging must name its target rather than let
+      # rsconnect infer one from the record. "connect-actions" is the nickname
+      # the action logs in under.
       - name: Deploy to Connect (dev)
         uses: posit-dev/connect-actions/deploy@main
         with:
@@ -25,7 +31,7 @@ jobs:
           content-guid: d7a36cae-8f27-448b-a478-61b81fbe3942
           path: pkg-r/inst
           github-token: ${{ github.token }}
-          rsconnect-args: --title biodiversity-explorer
+          rsconnect-args: --name connect-actions --title biodiversity-explorer
 
       - name: Deploy to Connect (staging)
         uses: posit-dev/connect-actions/deploy@main
@@ -34,4 +40,4 @@ jobs:
           content-guid: ad662e1b-5048-4acc-9ad7-f9478c92274e
           path: pkg-r/inst
           github-token: ${{ github.token }}
-          rsconnect-args: --title biodiversity-explorer
+          rsconnect-args: --name connect-actions --title biodiversity-explorer


jat255:
@simonpcouch I don't think this needs a review or anything, but FYI. It should fix #318

Issue #321, opened 2026-09-08
Inherited ellmer methods answer without the turn rules or a provenance marker


jat255:
`Commons` inherits from `ellmer::Chat` ([commons.R:213-215](https://github.com/posit-dev/commons/blob/d1833b2fdf2bbbe2d3ac37b345d914992cb320b5/pkg-r/R/commons.R#L213-L215)) and overrides only two of the six methods that can produce an answer. `chat()` and `stream_async()` are overridden. `stream()`, `chat_async()`, `chat_structured()` and `chat_structured_async()` are inherited without change.

This means that if a user constructs a `Commons` instance and uses one of those inherited methods, they don't get any of the commons-y goodness, since the turn rules never run. `stream()` and `chat_async()` skip the turn reminder, the restore reminder, and the conversation-turn span, which `chat()` installs at [commons.R:309-318](https://github.com/posit-dev/commons/blob/d1833b2fdf2bbbe2d3ac37b345d914992cb320b5/pkg-r/R/commons.R#L309-L318). They also skip the citation scan and the provenance aside, which only `stream_async()` adds. An answer from `commons$stream()` therefore reaches the reader with no provenance marker and no verified citations, and the trajectory record for that turn has no span.

This is probably worth consideration if we want to support those other methods, or (if not) should we prohibit them from being used in Commons? 

Two parts of this are worth separating:

- `stream()` and `chat_async()` produce prose for a person to read. These are the methods that need the turn rules.
- `chat_structured()` and `chat_structured_async()` return parsed data. A provenance marker appended to their output would make the JSON invalid, so the correct behavior for them is probably to abort with a message that names `stream_async()`, not to scan.

This is an artifact of inheriting directly from ellmer on the R side. The Python package instead composes an object from chatlas instead of inheriting from it, so we're using an object that we explicitly provide methods on. The underlying chatlas client is reachable only through the documented `client` attribute, which cleraly documents that it skips the scanner and the marker. Composition makes the marked path the only path. Inheritance makes it a subset of the surface, and the subset needs a new override every time ellmer adds a method.

This came up while building the Python chat adapter for the UI layer, and I figured it was worth opening an issue for consideration.

@simonpcouch

PR #322, opened 2026-09-08
feat(py): making `Commons` agent a subclass of chatlas.Chat

jat255:
`Commons` now [inherits directly from `chatlas.Chat`](https://github.com/posit-dev/commons/pull/322/files#diff-73bffe38e0d176e55662779d49a175bfe7797ad9fc85c1313f38b582b7ee04f8) rather than composing. This simplifies the interaction with shinychat and prevents needing to rely on lots of private internal methods to build a shinychat adapter (like in the first version of this PR).

## Why inherit

shinychat does not duck-type the object on its `client=` parameter, and the type (is it a `chatlas.Chat`?) gates both bookmarking and the history drawer. To pass those gates from the outside, an adapter has to reimplement chatlas's bookmark payload and shinychat's turn serialization, both against private upstream modules, which is a poor design. By inheriting from `chatlas.Chat`, we get those by default.

Inheriting also gives us a couple of other useful defaults: (1) a saved conversation records the provider and the model, because `client_info()` returns them for a chatlas client; and (2) turns move between clients as native `Turn` objects, which is what shinychat's own client-swap path appends to (rather than having to serialize to JSON).

This is similar to how the R side inherits from `ellmer::Chat`, so the two packages now agree. If future changes to shinychat implement duck-typing or adds a hooks for a non-chatlas one, we can revisit this, but I'm pretty satisfied with the design for now.

## What inheritance costs

An agent inherits chatlas's other answer-producing methods, and each of those submits a turn to the agent without  being wrapped by commons' turn handling, so they wouldn't get any of the provenance/citation stuff. This PR has those methods explicitly raises `NotImplementedError`: `chat_async()`, `stream()`, `chat_structured()`, `chat_structured_async()`, `extract_data()`, `extract_data_async()`, `to_solver()`, and `stream_async(data_model=...)`. `app()` and `console()` stay inherited, because they route back through `stream_async()` and `chat()`. There's currently a gap in this on the R side, which was previously opened as #321.

Copying a `Commons` agent is also not supported (and raises an error) because `Commons` agents can carry database connections in their data layer that cannot be copied. 

## Tests

[`test_shinychat.py`](https://github.com/posit-dev/commons/pull/322/files#diff-b1ad02952785a3f942279522e4b4e60d96815aa1da110d351557e9e33106af86) runs shinychat's own functions over an agent rather than asserting a described contract: its type guard, its turns adapter and title fallback, its client-swap turn handling, and its chatlas bookmark hooks in both directions. Driving a whole `shinychat.Chat` needs an active Shiny session, so that assertion belongs to the acceptance task (`f6e3`).

`pkg-py` is clean on `ruff check`, `pyrefly check src tests`, and 1330 tests.


cpsievert:
I'm okay with subclassing in the short term, especially if that makes your job the easiest without needing changes upstream. 

Longer term, I think we should consider both:

1. shinychat does a better job at duck typing
2. ellmer/chatlas agent gains hooks (or similar APIs) to support the needs of commons

For commons in particular, I think 2 would be the "dream" scenario. That is, a commons instance would become just a ellmer/chatlas instance with the proper middleware. That said, we also want 1 so that we have decent support for other "3rd party" backends like LangChain, etc.

PR #323, opened 2026-09-08
refactor: single root source for the browser assets

jat255:
## Summary

This PR moves the chat and viewer browser assets to a root `www/`, and `scripts/sync-shared.sh` regenerates the copy each package ships. This is M8 task 1, kata `jgxg`, and it unblocks the Python asset dependency and theme (`w4np`).

This approach follows the same source-plus-verified-copy pattern used elsewhere throughout the repo. Neither an R package nor a packaged Python wheel can read a file outside its own directory, so we have to do explicit copies. The CI failures are the guard that a file has gone stale or was modified in the wrong place.

## Verification

- `R CMD check` clean (0 errors, 0 warnings)
- R suite: 7341 passing with 23 environment-gated skips
- pkg-py: clean on ruff, pyrefly, and 1312 tests
- The staleness test was confirmed to fail against a hand-edited copy, not just to pass when synced

## R-side summary

@simonpcouch 

No R code changed. The only diff under `pkg-r/inst/www/` is the generated `README.md` the sync script adds; every asset is byte-identical to `main`, confirmed with `diff -r` against the root source.

The main thing that changes is that `pkg-r/inst/www/` is now generated output rather than the source location. Future changes to any assets should be done by editing the root `www/` and re-running the sync script instead; `test-assets-synced.R` is a new test that will fail when the two diverge.

One small caveat: the dependency's cache-busting stamp in `commons_chat_dependency()` uses `max(file.mtime(...))` to determine when to invalidate the browser's cache. Because our sync script uses `cp` (which refreshes mtimes), the version string for caching can change when the actual content did not. It's not really dangerous, but it may over-invalidates browser caches occasionally and cause re-downloads. (this is a very minor issue not worth fixing, IMO). To be "technically correct", we could swap the mtime-stamp behavior for a hash of the file contents instead, but I think that's overkill.

No hand-written R tests were deleted; `test-assets-synced.R` is new.


jat255:
The failing `check` job is not this branch's doing. Unmodified `main` fails the same test the same way ([run 34291834913](https://github.com/posit-dev/commons/actions/runs/34291834913)), filed as #324: a shinychat development build now refuses `chat$clear()` when conversation history is enabled. Every other job here passes, and the new asset test skips under `R CMD check` because the repository root is not reachable there, which is the intended branch.

Locally, with an older shinychat, the full R suite passes (7341 tests, 23 environment-gated skips) and `R CMD check` is clean apart from the pre-existing `glue` NOTE.

simonpcouch (approved):

PR #325, opened 2026-09-09
test: use public chat$new_chat() in restore-reminder test

### LINKED ISSUE #324: R CMD check is red on main: shinychat now refuses chat$clear() when history is enabled
`R CMD check` is failing on `main` with one test error:

```
Error ('test-chat.R:63:7'): commons_server queues a restore reminder when history is restored
Error in `chat$clear()`: Can't clear a chat with conversation history enabled. Use `chat$new_chat()` to start a new conversation.
```

CI resolves shinychat from r-universe (`shinychat_0.4.0.9000_dde163e`) rather than a pinned version, so a development build changed under us.

[shinychat #399](https://github.com/posit-dev/shinychat/pull/399) ("Add new_chat for conversation history"), commit [`dde163e`](https://github.com/posit-dev/shinychat/commit/dde163ea6b27099304a658f3649dd1fac98208bd), merged 2026-09-08 and added `new_chat()` along with a guard at the top of `client_clear()` in `pkg-r/R/chat_app.R`: when a history controller exists, `clear()` now aborts instead of clearing.

In commons, `shinychat::chat_server()` defaults to `history = TRUE`, and `commons_server()` calls it without passing `history` (`pkg-r/R/chat.R:116`), so there is no configuration reachable through commons in which `clear()` still works. `clear()` is not deprecated, it is closed off for the only setup this package creates. This needs a decision about the test rather than a rename.

The test uses `clear()` at line 63 to assert that the pending restore reminder is dropped, then separately asserts the same thing through `controller$new_chat()` two lines later.

Deciding what the test should assert is an R behavior question. The options look like: 
- exercise the clear path with history disabled, since that is now the only configuration where `clear()` is reachable
- or drop the `clear()` assertion because `new_chat()` already covers the same commons behaviour
- or keep asserting that commons drops the reminder and accept that upstream now owns which method gets there.

It's also worth a decision whether commons CI should pin the shinychat build it resolves, given upstream changes can cause CI failures unexpectedly (like this one).

@simonpcouch @cpsievert any thoughts?

cpsievert:
Closes #324

## Summary
shinychat (posit-dev/shinychat#399) made `clear()` abort when conversation history is enabled, which is the only configuration `commons_server()` creates — so the restore-reminder test's `chat$clear()` assertion had no reachable behavior left to exercise. This drops that assertion (the reset path is identical to `new_chat()`'s, and the agent-level lifecycle is already covered by unit tests in test-commons.R) and replaces the internal `controller$new_chat()` call with the public `chat$new_chat()` on the object `chat_server()` returns. One internal reach remains: firing a restore under `testServer` requires the history controller (`shinychat:::get_session_chat_bookmark_info`), since shinychat exposes no public restore trigger; a comment in the test says so.

## Verification
`testthat::test_file("tests/testthat/test-chat.R")` passes against shinychat main (0.5.0, local install from ../shinychat/pkg-r, which includes the `clear()` guard).

### DIFF
--- a/pkg-r/tests/testthat/test-chat.R
+++ b/pkg-r/tests/testthat/test-chat.R
@@ -53,6 +53,8 @@ test_that("commons_server queues a restore reminder when history is restored", {
       chat <- commons_server("chat", client = agent)
     },
     {
+      # shinychat has no public hook to fire a restore under testServer, so
+      # drive the history controller it stashes on the session.
       controller <- shinychat:::get_session_chat_bookmark_info(
         session,
         "chat.history-controller"
@@ -60,11 +62,7 @@ test_that("commons_server queues a restore reminder when history is restored", {
       controller$restore_app_state(list())
       expect_true(agent$.__enclos_env__$private$restore_reminder_pending)
 
-      chat$clear()
-      expect_false(agent$.__enclos_env__$private$restore_reminder_pending)
-
-      controller$restore_app_state(list())
-      controller$new_chat()
+      chat$new_chat()
       expect_false(agent$.__enclos_env__$private$restore_reminder_pending)
     }
   )


jat255:
LGTM, but I'd wait for @simonpcouch

simonpcouch (approved):

PR #327, opened 2026-09-09
feat(py): the chat asset dependency and commons.ui.theme()

jat255:
Adds the browser-asset dependency and the theme that carries it onto the page, the Python side of what `commons_theme()` does in R. kata `w4np`, M8 task 2. Stacked on #322.

## Summary

Adds `commons.ui` module to the Python side with `commons.ui.theme()` that builds the commons theme from chat Sass variables over `shinychat.page_chat_theme()` and adds an `HTMLDependency` from `www/commons-chat`. 

`commons.ui` is a real module now after this PR, so `import commons.ui` and `from commons.ui import theme` both work. A module-level lazy `__getattr__` on `commons` resolves it on first use, so importing `commons` defers importing shiny.

Files: [`_assets.py`](https://github.com/posit-dev/commons/pull/327/files#diff-daff9b4acf3e4ba73035e6f5afac7a341d377b599555f83e9ffc522244f3b5d6), [`_theme.py`](https://github.com/posit-dev/commons/pull/327/files#diff-eee2205ee368d321f2afdbcfade0fdd67bfa0114ae442f5f8d606d3a4bf66090), [`ui.py`](https://github.com/posit-dev/commons/pull/327/files#diff-b78bd1074561106d717d0cf918195c7447686a0032bee4c9ac26fde8b8d2e1f8).

## Implementation notes

@cpsievert, curious on your thoughts on this section...

Compared to R, py-shiny has no equivalent of `bslib::bs_bundle()` with `sass::sass_layer(html_deps=)`. `commons.ui.theme()` instead returns a `shiny.ui.Theme` subclass that overrides `_html_dependencies()`, which is shiny's a documented (maybe unintentionally?) point for this and has no public equivalent (see the comment on that private method [here](https://github.com/posit-dev/py-shiny/blame/main/shiny/ui/_theme.py#L595-L596). As an example, Shiny's own `ThemeBrand` overrides it the same way, to attach a fonts dependency. Even though that interface has been stable for 2 years, it's still a bit awkward to rely on a private internal method like this. Furthermore, there's an odd `__class__` rebind, which `theme()` needs because `shinychat.page_chat_theme()` builds a plain `Theme` internally.  An upstream public way for `Theme` to carry extra dependencies, or a `page_chat_theme()` that accepts the class it returns would be a cleaner way to deal with this.

## Shared fixture

The five Sass variables, the preset, and the behavior of a caller's own variable beating a commons default are cross-language, so they are pinned in [`tests/shared/chat-theme.json`](https://github.com/posit-dev/commons/pull/327/files#diff-1455706f051d50991a42f0acaf4e3f38c5927d7dffc9d79432f04ec144874266).

Only the Python suite runs those cases, which is @simonpcouch's call: reading the variables back through `bslib::bs_get_variables()` asserts little more than that bslib layered what it was handed, and asserting the theme reaches a page exercises shiny rather than commons. The fixture stays shared so the values stay recorded and an R runner can pick them up later. [`tests/shared/README.md`](https://github.com/posit-dev/commons/pull/327/files#diff-16141c05e0074e1636e6c6ea007d4a092fccb872b3aff67302e47f9b1d1258c3) says plainly that this is the one fixture there that a single suite runs.

`ruff check`, `pyrefly check src tests` and `pytest` pass in `pkg-py/`.

## R changes

@simonpcouch

No R source and no R tests change. The only R file this touches is the synced copy of the new shared fixture under `pkg-r/tests/testthat/fixtures/shared/`, which `scripts/sync-shared.sh` generates and CI checks for staleness. `commons_theme()` behaves exactly as it did, and no call site of it moves.


pkg-r/tests/testthat/test-theme.R
simonpcouch, line 1: I'm fine with dropping this entirely--these mostly test that the theme exists on our side, and are otherwise testing a shiny path that is not ours to test

simonpcouch (approved):
LGTM after dropping that one test!

cpsievert:
> it's still a bit awkward to rely on a private internal method like this.

Yea, agreed, but I'm ok with the hack in the short term if it works. Long term, let's add a more official API for this in shiny.

cpsievert (approved):

pkg-r/tests/testthat/test-theme.R
jat255: Removed!

PR #338, opened 2026-09-10
remove glue dependency

simonpcouch:


simonpcouch:
Resolves a NOTE on main.

PR #342, opened 2026-09-10
feat(py): commons.ui.app() for local development

jat255:
`commons.ui.app(agent)` is the complete app for local development, the page and theme and server composed the way `commons_app()` composes them. `demo.py` now serves its agent as a chat, as `pkg-r/inst/demo.R` does.

Stacked on #339, which adds the server function this builds on.

<details>
<summary>Detail (written by an agent)</summary>

**URL bookmarking became a history option.** R passes `enableBookmarking = "url"` to `shinyApp()`. `HistoryOptions(restore_mode="url")` on the chat is the same outcome, a conversation the address bar can link to, and shinychat documents it as needing no bookmark store. So `shiny.App(bookmark_store=)` keeps its `"disable"` default, `Chat.enable_bookmarking()` is not called, and R's `setBookmarkExclude()` has nothing left to exclude. An app that wants real Shiny bookmarks asks for `bookmark_store="server"` and `restore_mode="bookmark"` itself.

**The toolbar is a keyword.** R gates it on `rlang::is_interactive()`, which Python has no equivalent of. Inferring a deployment from environment variables would be a worse contract than an argument on a function whose stated purpose is local development.

**R's close button is left out.** It calls `shiny::stopApp()`, py-shiny has nothing equivalent, and `session.close()` ends one browser session while the server keeps running, which is different behaviour under the same label. The toolbar carries the dark-mode switch alone.

**One agent is shared by every session.** This is R's shape: `commons_app(client, ...)` takes an agent, not a factory, and the deployment path is `commons.ui.server()` with an agent built inside the server function. The review flagged the shared agent twice as a cross-session exposure risk, and I dismissed it rather than change a public API away from `pkg-r` on my own: if it is worth fixing, both packages should change together. The docstring and the README now say plainly that sessions share the agent. Worth your call.

**The demo keeps both entry points.** `shiny run demo.py` for the chat, `python demo.py` for the terminal, because `demo.ipynb` drives `ask()` and `tools_run()`. The greeting suggestions are derived from the same question list the terminal run asks, so the two cannot drift.

`ruff check`, `pyrefly check src tests` and the full suite pass with the `shiny` and `tracing` extras installed. The new tests cover the app's type, the restore mode its server wires, the assets and dark-mode switch on the page, the toolbar keyword, and the agent check.

</details>


jat255:
Noting a deferred issue from review: `commons.ui.app()`, like R's `commons_app()`, takes an agent object rather than a factory, so every session of the local-development app shares one agent and one conversation — two simultaneous viewers of a demo would see each other's conversation. This isn't a huge deal, but it's worth consideration at a later time. I filed it locally as kata `0szv` (related: `x0mt`) with three options to decide jointly for both packages: keep as-is and rely on the local-development documentation, accept a factory alongside an agent, or have the app always build its own from a factory.

PR #344, opened 2026-09-10
recover trajectories with omitted rich content

### LINKED ISSUE #341: trajectory reads are wonky with rich content in tool results
Namely, plots.

The plot tool result contains an ellmer::ContentImageInline. When ellmer records the next chat span, JSON serialization fails with `No method asJSON S3 class: S7_object`. ellmer suppresses that error and omits the span’s entire gen_ai.input.messages attribute.

Before an ellmer fix, we can probably represent that as missing content with a placeholder tool result in `trajectory_read()`. Probably doesn't make sense for ellmer to write the base64 (maybe? I don't know much about semconv), but it could at least write a placeholder tool result noting that a plot (or other rich content type) was there.

---

This might also be the case for rich content generally, e.g. if the user provides an image as part of their input.

simonpcouch:
Closes #341.

Reconstructs chat outputs from the last valid input snapshot within the same agent turn in `trajectory_read()`. Represents omitted rich inputs and tool results with explicit placeholders so affected conversations remain reviewable `trajectory_review()`.

### DIFF
--- a/pkg-r/R/trajectory-read.R
+++ b/pkg-r/R/trajectory-read.R
@@ -796,12 +796,30 @@ latest_chat_spans <- function(chat_spans) {
   for (span in chat_spans) {
     id <- span_conversation_id(span)
     prev <- latest[[id]]
-    if (is.null(prev) || span_time(span) > span_time(prev)) {
+    if (is.null(prev) || chat_span_is_preferred(span, prev)) {
       latest[[id]] <- span
     }
   }
 
   latest[order(vapply(latest, span_time, character(1)))]
+}
+
+chat_span_is_preferred <- function(candidate, current) {
+  candidate_is_title <- is_conversation_title_span(candidate)
+  current_is_title <- is_conversation_title_span(current)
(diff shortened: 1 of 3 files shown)


simonpcouch:
Works for me on staging.

Screenshot 2026-09-10 at 9 29 37 AM

simonpcouch:
Doesn't work for images in chat input yet.

simonpcouch:
Even text-only turns following a plot-containing turn will need to be substituted witha placeholder:

> ..the follow-up text is absent from the trace.
>
>Ellmer serializes the entire accumulated history into `gen_ai.input.messages`. The earlier plot makes that serialization fail, so ellmer omits the whole attribute—including the new text-only follow-up. It isn’t recorded elsewhere.
>
>Thus `trajectory_read()` can recover earlier exchanges, but only placeholder the follow-up. Proper recovery requires ellmer to serialize the offending rich result as a placeholder while retaining the rest of the messages.

simonpcouch:
Filed on ellmer: https://github.com/tidyverse/ellmer/issues/1145

PR #350, opened 2026-09-10
Situate measure `tool_result_display()`s with normal rich request display

### LINKED ISSUE #348: Situate measure `tool_result_display()`s with normal rich request display
Related to posit-dev/tlg-agent. That agent's measure return `shinychat::tool_result_display`s with:

```r
  ellmer::ContentToolResult(
    value = tidy_tlg(tt),
    extra = list(
      display = shinychat::tool_result_display(
        html = rtables::as_html(tt, width = "100%"),
        open = TRUE,
        full_screen = TRUE
      ),
      data = tt
    )
  )
```

This gives tlg-agent an entry point to show rich HTML tables and provide the model a data frame in the agent's R session.

`tool_result_display(show_request)` defaults to TRUE, so the tool UI looks like this:

> Image

For measure results, commons should situate `tool_result_display` in the otherwise-default measure UI, which shows the arguments and measure descriptions:

> Image

simonpcouch:
Closes #348. More context in that issue.

The app with this PR shows the measure title and description, where the description fades out in the third line with a 'See more' button, so that users can focus on the main tool result display:

> Screenshot 2026-09-10 at 3 50 37 PM

### DIFF
--- a/pkg-r/R/tools.R
+++ b/pkg-r/R/tools.R
@@ -323,7 +323,8 @@ tool_call_measure <- function(private) {
         injections = private$injections,
         handles = private$handles,
         sources = private$sources,
-        measure_provenance = private$measure_provenance
+        measure_provenance = private$measure_provenance,
+        measure_display = private$measure_display
       )
     },
     paste(
@@ -501,7 +502,8 @@ call_measure_tool <- function(
   injections = list(),
   handles = NULL,
   sources = list(),
-  measure_provenance = list()
+  measure_provenance = list(),
+  measure_display = list()
 ) {
   td <- registry[[name]]
   if (is.null(td)) {
@@ -515,6 +517,7 @@ call_measure_tool <- function(
   footer <- measure_source_footer(
     measure_provenance[[name]] %||% character()
   )
+  metadata <- measure_metadata(td, measure_display[[name]])
   args <- validate_measure_args(td, parse_json_args(arguments))
   # A measure takes a source's connection by the source's name; a board source
   # must have its pins loaded before that connection can answer a query.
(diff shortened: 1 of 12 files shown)


www/commons-chat/commons-chat.css
```diff
@@ -404,6 +404,73 @@ shiny-chat-container
...
 }
 
+.commons-measure-description-summary {
+  display: -webkit-box;
+  overflow: hidden;
+  -webkit-mask-image: linear-gradient(to bottom, #000 55%, transparent);
+  mask-image: linear-gradient(to bottom, #000 55%, transparent);
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 3;
+}
```
simonpcouch, line 414: The built-in measure tool result display previously showed all of the documentation for the measure before the measure result. Now, we show 200-ish characters and allow the user to click to see more.

pkg-r/R/tools.R
```diff
@@ -577,12 +599,21 @@ measure_content_tool_result <- function(td, result, handles, footer) {
...
       icon = icon,
-      footer = footer
+      footer = footer,
+      show_request = FALSE
     )
   } else if (is.list(display)) {
     display$title <- display$title %||% title
     display$icon <- display$icon %||% icon
     display$footer <- display$footer %||% footer
+    display$show_request <- FALSE
```
jat255, line 609: Is there a situation where we'd want to respect a `show_request` coming in as `TRUE`?

Should we instead do `display$show_request <- display$show_request %||% FALSE` 

(I don't fully grasp where `display` comes from, so feel free to ignore this...)

www/commons-chat/commons-chat.css
```diff
@@ -404,6 +404,73 @@ shiny-chat-container
...
+.commons-measure-metadata-item:has(.commons-measure-details[open])
+  > .commons-measure-description-summary {
+  display: none;
+}
+
+.commons-measure-details {
+  color: var(--bs-secondary-color, #6c757d);
+}
+
+.commons-measure-details > summary {
```
jat255, line 425: Kimi pointed out a consistency issue that this new summary style has no `:focus-visible` rules, unlike both sibling interactive components in the same file (`.commons-provenance-info-trigger:focus-visible` at line 150, `.commons-run-r-details > summary:focus-visible` at line 591).

pkg-r/R/tools.R
```diff
@@ -1089,25 +1160,51 @@ measure_metadata_html <- function(metadata) {
...
+      details <- metadata_details[[i]]
+      details <- if (!nzchar(details)) {
+        ""
+      } else {
+        sprintf(
+          paste0(
+            "<details class=\"commons-measure-details\">",
+            "<summary>",
+            "<span class=\"commons-measure-details-more\">See more</span>",
+            "<span class=\"commons-measure-details-less\">See less</span>",
```
jat255: Accessibility nit: It may be worth adding an `aria-label` naming the measure here? If you had multiple measures, a screen reader would show the same "Show more"/"Show less" without context

jat255 (approved):
Minor comments. I'll send a follow-up PR stacked on this one to make the corresponding python changes

EDIT: anything to add in `pkg-r/inst/demo.R` to illustrate this change?

www/commons-chat/commons-chat.css
simonpcouch: You got it!

pkg-r/R/tools.R
simonpcouch: Because `display$show_request` defaults to TRUE, I think that most users just wouldn't know about it or change it and thus would never see this feature. I'm personally fine with overwriting it for now.

PR #351, opened 2026-09-10
feat(py): line-delimited JSON protocol and Arrow value transfer

jat255:
This PR adds the wire between the code-execution driver and its worker: one JSON object per line, with values crossing by copy as Arrow IPC for frames and JSON for scalars. A frame that would cost the driver more than the channel allows is refused before it decompresses. Nothing consumes it yet, since the worker and the driver are separate units of work.

Pickling is forbidden in either direction, because it is inherently insecure. Loading a pickle runs whatever opcodes it is handed, and on the way back those opcodes were written by whatever the worker just ran. 

<details>
<summary>Design decisions and review notes (agent-written)</summary>

**Arrow rides base64 inside the JSON line.** The alternative was a second binary channel, but the `exec`-shaped backend offers stdin, stdout and stderr and nothing else, and a container-hosted backend may not pass extra file descriptors. One channel is what keeps a Connect backend a drop-in.

**Frames arrive as whichever library sent them.** The payload names the library and the far end rebuilds with it. The driver and the worker share an interpreter and site-packages, so it is always available. `commons` accepts pandas and polars everywhere else without preferring one, and the transport should not quietly impose a choice on the model.

**A value that cannot cross becomes an `OpaqueValue` carrying its repr.** An open connection or a fitted model has no copy. Refusing outright would fail a whole call over a value the model may not want, and a repr is what a REPL shows. Frames are not exempt: a column Arrow cannot hold costs the value rather than the call.

**The frame cap counts what pandas will hold, not what Arrow sends.** A null column writes no buffer at all, so a wide frame of them, or a single row of `list<null>`, is kilobytes on the wire and a pointer per cell once pandas has it. The schema is the only part of a stream that says which fields write nothing. The pre-flight reads it, then charges each such field by the length its own node declares. Only pandas pays. Arrow holds a null column as a length and polars keeps it that way. Charging a frame bound for either would refuse what this module is willing to encode.

**Nothing crosses the frame boundary as its own exception.** The far end chooses the bytes and the library name separately, so a table Arrow accepts can still be one pandas or polars will not hold. A duplicate column name is the everyday case for polars. Those refusals arrive as `ProtocolError`, like any other malformed payload.

**JSON nesting depth is bounded by the protocol, not the interpreter.** Before 3.14, `json.dumps` and `json.loads` refused deep nesting with `RecursionError`; the rewritten parser and encoder are iterative and never will, while everything a value meets after the codec — `repr`, re-serialization, display — still recurses. `_JSON_DEPTH_LIMIT` keeps the codec the single chokepoint: a value nested past it crosses as its repr, and a line nested past it is refused. The check is a stack of iterators, so its memory tracks depth rather than breadth — a wide hostile line cannot exhaust the driver between parse and refusal.

**`STREAM_LIMIT` is exported and has to be used.** `asyncio` gives a stream 64 KiB by default, which any real frame exceeds, so the worker lifecycle task has to pass it to `create_subprocess_exec`. Worth naming now rather than meeting as a mysterious failure on the first result that matters.

**The module imports standalone, without `commons`.** The worker runs under `-I` with only its own directory on the path, and has no business holding the agent's dependencies. A subprocess test pins this using the environment allowlist from #251.

No `tests/shared/` fixture: the R package uses `callr`'s own transport, so this protocol is not a cross-language contract.

Every commit was reviewed by roborev, and every finding is fixed or dismissed with the reasoning recorded on the job. Three are worth a reviewer's attention:

- A pandas column of complex numbers raises `ArrowNotImplementedError`, which is not a `ValueError` and escaped the fallback.
- An object array's `tolist()` bypassed the JSON depth check.
- The size cap read `nbytes`, which Arrow reports as zero for a null column however many cells it holds.

Two findings were dismissed with evidence on the job, 608 and 737. Each refusal the cap adds has a test that fails when the code it covers is disabled.

</details>


pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,306 @@
...
+    "encode_value",
+    "read_message",
+    "write_message",
+]
+
+# How long a single message line may be. A frame crossing as base64 Arrow is
+# nothing like the 64 KiB `asyncio` allows a stream by default, and a driver
+# that leaves the default in place will fail on the first real result rather
+# than on an unusual one. Whoever opens the channel passes this as `limit=`.
+STREAM_LIMIT = 64 * 1024 * 1024
```
jat255, line 41: The worker lifecycle task has to pass this to `create_subprocess_exec`; the 64 KiB default would fail on the first frame-shaped result.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,306 @@
...
+    their JSON shape: a tuple comes back a list, and a dictionary's keys come
+    back as strings.
+    """
+    library = _frame_library(value)
+    if library is not None:
+        data = _to_arrow_ipc(value, library)
+        # A column Arrow cannot hold — objects in a pandas `object` column,
+        # or complex numbers. Being a frame is no reason to be exempt from
+        # the fallback: one bad column should cost the value, not the call.
+        if data is None:
```
jat255, line 145: Judgment call worth a second opinion: a frame with one column Arrow cannot hold degrades to a repr rather than failing the call, on the grounds that the model may not have wanted the value.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,306 @@
...
+            )
+    except (KeyError, TypeError, ValueError) as error:
+        raise ProtocolError(f"malformed value payload: {error}") from error
+    raise ProtocolError(f"unknown value encoding: {encoding!r}")
+
+
+# Recognized without importing anything: a value can only be a pandas frame
+# if pandas is already imported, and forcing the import here would make the
+# worker pay for a library the code it ran never asked for.
+def _frame_library(value: Any) -> str | None:
```
jat255, line 187: Deliberately not an import: a value can only be a pandas frame if pandas is already imported, so this recognizes frames without making the worker pay for a library its code never asked for.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,862 @@
...
+    try:
+        text = repr(value)
+    except Exception:  # noqa: BLE001 - the last resort may not itself raise
+        return f"<{type(value).__name__} whose repr raised>"
+    if len(text) > _REPR_TEXT_LIMIT:
+        return f"{text[:_REPR_TEXT_LIMIT]}… [{len(text):,} characters in full]"
+    return text
+
+
+def _exceeds_json_depth(value: Any, limit: int = _JSON_DEPTH_LIMIT) -> bool:
```
jat255: Review finding, fixed in 31fe2c2: the first version extended the stack with one tuple per child, so a stream-sized line of flat empty lists could exhaust the driver between parse and refusal. The stack now holds one iterator per level, and memory tracks depth rather than breadth.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,862 @@
...
+    else:
+        numpy = sys.modules.get("numpy")
+        if numpy is None or not isinstance(value, numpy.ndarray):
+            return _FALL_BACK
+        try:
+            candidate = value.tolist()
+        except Exception:  # noqa: BLE001 - any failure means the repr fallback
+            return _FALL_BACK
+    try:
+        # An object array's tolist can hide arbitrarily deep nesting; the
```
jat255: Review finding, fixed in 31fe2c2: a 0-d object array's tolist() hands back the object it holds, which bypassed the depth check, crossed as JSON, and sank the whole message at the far end. The coerced candidate is now checked before it may cross; test_an_object_array_holding_a_deep_value_crosses_as_text pins it.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,1055 @@
...
+]
+
+# The longest allowed message line. Whoever opens the channel must pass this
+# as `limit=`: `asyncio` defaults a stream to 64 KiB, which any real frame
+# exceeds. A message past the limit is a `ChannelError` from `read_message`.
+STREAM_LIMIT = 64 * 1024 * 1024
+
+# The largest a frame may be after Arrow decompression. IPC bodies can be
+# compressed, so a line well under STREAM_LIMIT can decode to far more.
+# 1 GiB covers any real frame while bounding what a hostile worker can make
```
jat255: The worker lifecycle task has to pass this to `create_subprocess_exec`, since the 64 KiB default fails on the first frame-shaped result.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,1055 @@
...
+    if frame is not None:
+        library, frame_value = frame
+        data = _to_arrow_ipc(frame_value, library)
+        # A column Arrow cannot hold (objects in a pandas `object` column,
+        # complex numbers) or a frame too large for the channel falls back
+        # to repr: one bad column costs the value, not the call.
+        if data is not None and len(data) <= _FRAME_WIRE_LIMIT:
+            return {
+                "encoding": "arrow",
+                "library": library,
```
jat255, line 234: Judgment call worth a second opinion: one column Arrow cannot hold costs the value rather than the call, on the grounds that the model may not have wanted it.

pkg-py/src/commons/_execution/_protocol.py
```diff
@@ -0,0 +1,1055 @@
...
+    # Only pandas spends a pointer on a null value. Arrow holds a null
+    # column as a length and polars keeps it that way, so the cells the
+    # caps below charge for are free unless the frame becomes a pandas
+    # one, and charging them anyway would refuse a frame this module is
+    # willing to encode.
+    cells = library == "pandas"
+    try:
+        _check_ipc_size(data, max_frame_bytes, cells)
+        reader = pyarrow.ipc.open_stream(pyarrow.py_buffer(data))
+        table = _read_capped(reader, pyarrow, max_frame_bytes, cells)
```
jat255, line 460: Judgment call worth a second opinion: only pandas is charged for null cells, because Arrow and polars hold a null column as a length (measured at 200M rows, ~0 bytes), and charging them would refuse frames this module itself encodes.

pkg-py/src/commons/_execution/_protocol.py
jat255: Confirmed: this matches the module's documented invariant — a bad value costs the value, never the call. Coercing the column would silently change the model's data, and raising would break the never-raise contract. Keeping as-is.

pkg-py/src/commons/_execution/_protocol.py
jat255: Confirmed: the charge models real decoded memory. Arrow and polars hold a null column as a length, so charging them would refuse frames this module encodes itself. Keeping as-is.

Issue #346, opened 2026-09-10
document that Connect dependency for trajectory logging is >2026.09.0


simonpcouch:

simonpcouch:
Well, greater than or equal to.

PR #359, opened 2026-09-11
document Connect version and warn when `commons(log = TRUE)` and Connect < 2026.09.0

simonpcouch:


simonpcouch:
Ope, closes #346.

PR #365, opened 2026-09-12
add a shared R/Python site, move R site to /r/

simonpcouch:
My goals with this site were:

* Should not feel vibe-coded.
* Make it easy as possible to get to the R/Python documentation quickly
* In terms of feel, be halfway between Quarto / a scientific document and a modern site. 

https://github.com/user-attachments/assets/1356a962-c2b5-4de8-af09-bc44a3e7d58d

The content is essentially the announcement post without the "Announcing" language. The "Python" button links to `/pkg-r` on GitHub, for now.


jat255 (approved):
LGTM! I have a slight worry about the hand-built HTML getting out of sync at some point, so at some point I think we should have an action that builds pkgdown, quartodoc for python (when those docs exist), and then the main page. I think this is fine for now though.

skaltman (approved):
looks good to me! I do think we should sync all the trust flow diagrams to use the new one that you made. It makes more sense. But that doesn't have to happen here.

PR #366, opened 2026-09-12
docs(py): add a documentation site for the Python package

jat255:
This PR adds a documentation site for the Python package at `/py/`: a quartodoc API reference, an introduction page,  a security and governance page, and a page describing the feature parity with the R implementation. The landing page's Python button now points there rather than at the package README on GitHub.

You can see what these look like from the nifty new doc preview action from #370: 

- https://posit-dev.github.io/commons/pr-370/py/
- https://posit-dev.github.io/commons/pr-370/r/

## R changes

The changes are only in the documentation  — no package code changes with this PR.

- `_pkgdown.yml` gains a navbar item on the right that links to the new Python site at `https://posit-dev.github.io/commons/py/`
- `extra.css` puts the R logo before the package name in the navbar via a `.navbar-brand::before` rule, matching how the Python site shows its logo, with the two logo SVGs added under `pkgdown/assets/logos/` via the root-level asset syncing described in the next section 

The logos and figures are handled by the existing root-artifact syncing:

- A new root-level `logos/` directory is the single source for the two language marks, and `scripts/sync-shared.sh` copies it into `pkg-py/docs/assets/logos/` and `pkg-r/pkgdown/assets/logos/`; the Python site's provenance marker SVGs come from the already-shared `www/commons-chat/figs`, copied into `pkg-py/docs/assets/figs/`.
- `verify-shared-synced.yaml` picks up the new source and destination paths so CI fails on a stale copy.

### Screenshots

**Python home page**

Note the python logo in the top left, and a link to the R version of the package on the top right

image

**API Reference built with quartodoc**

image

**R homepage, showing R logo in top left and link to python package in top right**:

image

<details>
<summary>Agent-written details</summary>

Refs kata 3w6z.

The pages cover the ground the R vignettes cover, and leave out what the Python package does not have rather than describing the R behaviour: no code-execution row in the provenance-outcome table, no warehouse semantic layer section, no agent skill or trajectory review.

That matters most on the governance page. No tool runs model-written Python, so the sandboxing, network, and subprocess material has no counterpart, and the page says so plainly instead of leaving the question open. Going the other way, the SQL check parses with sqlglot and allowlists read-only statement shapes, so it catches writes hidden inside a read that a first-word test admits.

Two claims drafted from R behaviour were wrong for Python and were corrected before filing: definitions compile only for DuckDB, Snowflake, and Databricks, and the guard's allowlist is wider than `SELECT` plus the set operations.

The two deploy workflows now share one non-PR concurrency group, because both push to `gh-pages` and a release triggers both.

Verified: ruff, pyrefly, and the 1499 pytest cases pass; the site builds clean from a fresh venv using the workflow's own install command; all eight external links and three internal anchors resolve.
</details>


simonpcouch (approved):
Nice, looks good!

FYI, the pkgdown Action to publish the _production_ version of https://posit-dev.github.io/commons/r/ (e.g. 0.1.0) will not push to the main site after merging here since the R DESCRIPTION version is now dev 0.1.0.9000. It's a bit fiddly to get this right; I can make it happen after merge.

PR #368, opened 2026-09-13
ci: publish a documentation preview for each pull request

jat255:
This PR adds a documentation preview for each pull request. See the comments below for examples. It publishes the landing page, the R site, and the Python site as one tree under `pr-<number>/` on the `gh-pages` branch, and then the action posts a comment with a link to the deployed preview (see below). There's also an "auto-cleanup" action that runs when a PR is merged (which hasn't been tested yet, but we can adjust as needed).

The PR also adds [`scripts/preview-docs.sh`](https://github.com/posit-dev/commons/pull/368/files#diff-e23b0136a4e9d4af4833bbc72cadbd8c41a4532ffd097056f0ba8d2d8e93c209) which can be used to build and serve the docs for both packages locally.

The other change (which probably should have gone on #366) is to give all three sites one shared favicon.

Stacked on #366.

<details>
<summary>Implementation notes (agent-written)</summary>

The rationale for each piece is in a comment at the top of the file, so this only says where to look.

**Publishing.** `pkgdown.yaml` and `quartodoc.yaml` gain a preview deploy alongside the existing one. [`noindex-preview.py`](https://github.com/posit-dev/commons/pull/368/files#diff-5ea4a0709189b5a93df0eed78516720e7b727cac325079a9ab2b81d4a3f9fdb7) adds a `robots` meta tag to every page, because a `robots.txt` under `posit-dev.github.io` is never read. [`link-r-dev.sh`](https://github.com/posit-dev/commons/pull/368/files#diff-3aaa87bd4aaeb16db45ac1343e158619f55e3c2c22d0b6adae283a28db9a0751) redirects `r/` to `r/dev/`, which is where pkgdown puts a development version. It runs for previews only, and the file says why moving the tree instead would break site search.

**Cleanup.** [`preview-cleanup.yml`](https://github.com/posit-dev/commons/pull/368/files#diff-5223b65676c47b22e92043ab7a17c03ad70f3b10a8112a2d8241e2727ebff431) removes a preview when its pull request closes, and a weekly schedule collects whatever the close event missed. The schedule cannot run until this merges, so the manual run defaults to a dry run.

**Forks.** A fork gets no preview, because building one would serve unreviewed HTML from `posit-dev.github.io`. [`preview-fork-notice.yml`](https://github.com/posit-dev/commons/pull/368/files#diff-dcda55300fec46672a65648b12712965c0353d416e5d3bc48a67978c49c56950) says so in a comment. It uses `pull_request_target` because a fork's token cannot write one; it checks out no code and runs none.

**Favicon.** `favicon/` at the repository root is the source, and [`sync-shared.sh`](https://github.com/posit-dev/commons/pull/368/files#diff-5abf301668aa8a6bcb4e8de21898196681b51bdc634b05f7ccbe45cf8c7ccc53) copies it into the two packages and the landing page. Destinations marked `:bare` skip the generated README, because their contents go onto a website as they stand.

**Tests.** `.github/scripts/` belongs to no package, so `scripts-check.yaml` runs `test_noindex_preview.py` on it.

**Verification.** `scripts/preview-docs.sh` built both sites and served them; the landing page, `r/dev/`, and `py/` all resolved, and their links to each other worked. `scripts/sync-shared.sh` regenerates every destination with no drift.

</details>

## R changes

Favicon assets are now copied by `scripts/sync-shared.sh` into `pkg-r/pkgdown/favicon/`, which pkgdown picks up when it builds the site. 

One little wrinkle/oddity is the interaction with pkgdown's automatic development mode. Because the deployed site expects the R site at `/r`, but pkgdown writes into `/r/dev`, the `link-r-dev.sh` script was necessary to provide a shim that writes a redirect into `r/index.html` only when that file does not exist (this is true for a development version and false for a released one). If the R package ever builds a preview from a released version, the redirect correctly does not appear. I'm not overly familiar with the benefits of pkgdown's dev mode, but I'm wondering if it's causing more trouble than value here.


jat255:
For posterity of what happened here, I accidentally merged this branch into the PR #366 branch prematurely, and then Github auto-closed this PR, which I couldn't re-open. #370 takes its place.

PR #370, opened 2026-09-13
ci: publish a Python documentation preview for each pull request

jat255:
This PR adds a documentation preview for each pull request that rebuilds the Python site. It publishes the landing page and the Python site as one tree under `pr-<number>/` on the `gh-pages` branch, and then comments a link to it. A cleanup action removes the tree when the pull request closes.

Per @simonpcouch's request, the R site does not get a preview build, and keeps the usual R package arrangement: `r/dev/` holds the development version (i.e. `main`), `r/` holds the released documentation, and `r/` is redeployed by hand at release.
 
The PR also adds a helper script, [`scripts/preview-docs.sh`](https://github.com/posit-dev/commons/pull/370/files#diff-e23b0136a4e9d4af4833bbc72cadbd8c41a4532ffd097056f0ba8d2d8e93c209) to build and serve both sites locally, and gives all three sites one shared favicon.

Stacked on #366. This replaces #368, which was accidentally merged and then unwound.

<details>
<summary>Implementation notes (agent-written)</summary>

The rationale for each piece is in a comment at the top of the file, so this only says where to look.

**Publishing.** [`quartodoc.yaml`](https://github.com/posit-dev/commons/pull/370/files#diff-07a2fba81e09d57a1b44944153af599175d8841aa7a8204e51ac1adfebeff57e) gains a preview deploy alongside the existing one. [`noindex-preview.py`](https://github.com/posit-dev/commons/pull/370/files#diff-5ea4a0709189b5a93df0eed78516720e7b727cac325079a9ab2b81d4a3f9fdb7) adds a `robots` meta tag to every page, because a `robots.txt` under `posit-dev.github.io` is never read. [`pkgdown.yaml`](https://github.com/posit-dev/commons/pull/370/files#diff-eb3d51fe47cf000aab65c6172c06d12a34ae765221529a05451295d5e9149e80) still builds the R site on a pull request, as a check that it builds, but deploys nothing there.

**Cross-site links.** A preview contains the landing page but no R site, so the links into the R site are absolute in [`docs/index.html`](https://github.com/posit-dev/commons/pull/370/files#diff-b04b38d4e36f7a7171aeb211bf933eaf36d41d9866ebbc3639f673f84dc350ae) and [`_quarto.yml`](https://github.com/posit-dev/commons/pull/370/files#diff-65d621b5c46ce4afce99fc13c25419a8c16462f1e2bbc1090271ce09880bc9ba). They point at the published site from everywhere. This matches what [`_pkgdown.yml`](https://github.com/posit-dev/commons/pull/370/files#diff-5e1c8b5cbe88768a7f2cb75aed5e241c2815514ea336b411c38f85e371aeacd9) already does for its Python link, which has to survive being served from both `r/` and `r/dev/`.

**Cleanup.** [`preview-cleanup.yml`](https://github.com/posit-dev/commons/pull/370/files#diff-5223b65676c47b22e92043ab7a17c03ad70f3b10a8112a2d8241e2727ebff431) removes a preview when its pull request closes, and a weekly schedule collects whatever the close event missed. The schedule cannot run until this merges, so the manual run defaults to a dry run.

**Forks.** A fork gets no preview, because building one would serve unreviewed HTML from `posit-dev.github.io`. [`preview-fork-notice.yml`](https://github.com/posit-dev/commons/pull/370/files#diff-dcda55300fec46672a65648b12712965c0353d416e5d3bc48a67978c49c56950) says so in a comment. It uses `pull_request_target` because a fork's token cannot write one; it checks out no code and runs none.

**Favicon.** `favicon/` at the repository root is the source, and [`sync-shared.sh`](https://github.com/posit-dev/commons/pull/370/files#diff-5abf301668aa8a6bcb4e8de21898196681b51bdc634b05f7ccbe45cf8c7ccc53) copies it into the two packages and the landing page. Destinations marked `:bare` skip the generated README, because their contents go onto a website as they stand.

**Tests.** `.github/scripts/` belongs to no package, so `scripts-check.yaml` runs `test_noindex_preview.py` on it.

**Verification.** `scripts/preview-docs.sh` built both sites and served them. The landing page, `r/dev/`, and `py/` all resolved, and their links to each other worked. `scripts/sync-shared.sh` regenerates every destination with no drift.

</details>

## R changes

Two things change on the R side:

Favicon assets are now copied by `scripts/sync-shared.sh` into `pkg-r/pkgdown/favicon/`, which pkgdown picks up when it builds the site. The only visible difference is that the R site serves the same icons as the other two.

[`pkgdown.yaml`](https://github.com/posit-dev/commons/pull/370/files#diff-eb3d51fe47cf000aab65c6172c06d12a34ae765221529a05451295d5e9149e80) has a few changes to manage concurrency between itself and `quartodoc.yaml`, so race conditions between different builds don't clobber the status of the `gh-pages` branch.


.github/workflows/pkgdown.yaml
```diff
@@ -52,3 +57,75 @@ jobs:
...
+      # fork-authored HTML to posit-dev.github.io means serving unreviewed
+      # content from our own origin.
+      # Preview only. The production deploy above publishes r/ as the released
+      # documentation, and this would replace its index with a redirect.
+      - name: Point r/ at the development site
+        if: >
+          github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.fork != true
+        working-directory: .
+        run: .github/scripts/link-r-dev.sh docs
```
simonpcouch: re:

> One little wrinkle/oddity is the interaction with pkgdown's automatic development mode. Because the deployed site expects the R site at /r, but pkgdown writes into /r/dev, the link-r-dev.sh script was necessary to provide a shim that writes a redirect into r/index.html only when that file does not exist (this is true for a development version and false for a released one). If the R package ever builds a preview from a released version, the redirect correctly does not appear. I'm not overly familiar with the benefits of pkgdown's dev mode, but I'm wondering if it's causing more trouble than value here.

It does provide a good bit of value here, and I'd prefer we keep it as-is. It prevents accidentally changing the production version of the website unless you've actually sent a new version to CRAN—we assume that most people use the CRAN version, so we keep the site pinned at that version. I can make the update-released-site-with-a-dev-version change happen after merge of the other PR, but I'd prefer we keep this as-is. Otherwise, we'll get a `/dev` version of the site when we merge to main, but it's personally not important to me to have deployed documentation previews per-PR.

